PR #10, opened 2015-10-28
added ADAM SGD

cbecker:
This adds the ADAM SGD algorithm.

I am decaying the rate manually in the ADAM functions, as I was not sure how to integrate it properly with the other MXNet learning rate functions.

### DIFF
--- a/src/optimizer.jl
+++ b/src/optimizer.jl
@@ -30,3 +30,4 @@ end
 
 
 include("optimizers/sgd.jl")
+include("optimizers/adam.jl")
--- a/src/optimizers/adam.jl
+++ b/src/optimizers/adam.jl
@@ -1,0 +1,64 @@
+
+@defstruct ADAMOptions Any (
+  (lr           :: Real = 0.001, lr > 0),
+  (lr_decay     :: Real = 1.0, lr_decay > 0),
+  (beta1        :: Real = 0.9,  beta1 > 0),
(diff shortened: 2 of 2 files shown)


cbecker:
also, I am not sure if the parameters (beta1Power, beta2Power, etc) should be Float64 or Float32.

pluskid:
@cbecker Thanks! The learning rate and decay looks fine. I will do some refactoring later on to make use of existing learning rate scheduler (the current interface is not complete yet, so decay is not supported).

Before I merge this, could you explain the choice of `Float32` for `beta1Power`, `beta2Power`? Thanks again!

cbecker:
Actually they should be Float64, I am not sure why I wrote Float32. I will fix it now.

Which machine precision is used inside MXNet? I suppose weights are converted to float, to speed up computations (and on GPUs because most GPUs do not work well with doubles)

pluskid:
@cbecker Internally, Float32 is used for tensors and arrays. But for scalar parameters, anything should be fine though. I was just wanting to make sure that I did not miss some special requirement that is needed for the ADAM solver.

cbecker:
Ok I see, no problem, then it should be fine now. You can merge.
As side note, in my experiments ADAM converges much faster than SGD, even with the default params.

pluskid:
Thanks! Oh, Great! I will try adam later!

Issue #15, opened 2015-11-04
mx._Native passing Ptr{Void} as parameter


vchuravy:
I am working on a basic interface to use Julia functions to implement operators. If I see it correctly `mx._Native` is the operator used to create Native Operators and its expects a `Ptr{Void}` as info parameter [1]. That void pointer should point to a struct containing function pointers.

````julia
immutable NativeOpInfo
         forward :: Ptr{Void}
         backward :: Ptr{Void}
         infer_shape :: Ptr{Void}
         list_outputs :: Ptr{Void}
         list_arguments :: Ptr{Void}
       
         function NativeOpInfo(forward :: Function, backwards :: Function, infer_shape :: Function, list_outputs :: Function, list_arguments :: Function)
           f = cfunction(forward,   Void, (Ptr{Ptr{Cfloat}}, Ptr{Cint}, Ptr{Cuint}, Ptr{Cint}))
           b = cfunction(backwards, Void, (Ptr{Ptr{Cfloat}}, Ptr{Cint}, Ptr{Cuint}, Ptr{Cint}))
           i = cfunction(infer_shape, Void, (Cint, Ptr{Cint}, Ptr{Ptr{Cuint}}))
           lo = cfunction(list_outputs, Void, (Ptr{Ptr{Ptr{Cchar}}},))
           la = cfunction(list_arguments, Void, (Ptr{Ptr{Ptr{Cchar}}},))
           
           new(f, b, i, lo, la)
         end
       end

````

After creating the appropriate object I want to pass it as a Ptr to mx_.Native

````julia
mx._Native(info = pointer_from_objref(info), name = :test)
ERROR: MXNet.mx.MXError("Invalid Parameter format for info expect  but value='Ptr{Void} @0x00007fe01e46c2b0'")
 [inlined code] from /home/wallnuss/.julia/v0.4/MXNet/src/base.jl:57
 in ___Native#13__ at /home/wallnuss/.julia/v0.4/MXNet/src/symbol.jl:56
````

I was thinking that maybe `Ptr{Void}` are handled differently in MXNet.jl?

[1] [native_op-inl.h](https://github.com/dmlc/mxnet/blob/e4ccd2b4ebd180636d448f7128977b16da5040df/src/operator/native_op-inl.h#L25)

vchuravy:
So it seems that the pointer is passed as a string from [1] to [2] ...

In Python
```py
>>> class NumpyOpInfo(Structure):
...   _fields_ = [
[...]

PR #16, opened 2015-11-05
Base structure for Native Operators

vchuravy:
Thanks to the help in #15 I managed to get past the initialization stage of _Native.

With the test_enty functions defined in src/nativeops.jl
````
info = mx.NativeOpInfo(fb_entry, fb_entry, infer_entry, list_entry, list_entry)
pstring = bytestring("0x", hex(reinterpret(UInt, pointer_from_objref(info))))
mx._Native(name = :test, info = pstring)
````

Works without causing a seqfault :tada:

Todo:
- [ ] Defining API exposed to Users
- [ ] Defining (gc-safe) entry points (maybe also thread safe?)

### API
I was thinking of something along the lines of
````
immutable Softmax <: mx.NativeOp end

function forward(::Type{Softmax}, in_data, out_data)
  ...
end

function forward(::Type{Softmax}, out_grad, in_data, out_data, in_grad)
  ...
end

function infer_shape(::Type{Softmax},  in_shape)
  ...
end

function list_output(::Type{Softmax})
  ...
end

function list_output(::Type{Softmax})
  ...
end
````

### DIFF
--- a/src/nativeops.jl
+++ b/src/nativeops.jl
@@ -147,7 +147,7 @@ type NDArrayOpInfo
     # The return value of the callbacks is stored in _FB and the first element is the
     # libuv handle.
     r_forward = Ref(_FB(cb_f.handle))
-    r_backward = Ref(_FB(cb_f.handle))
+    r_backward = Ref(_FB(cb_b.handle))
 
     p_f = Base.unsafe_convert(Ptr{Void}, r_forward)
     p_b = Base.unsafe_convert(Ptr{Void}, r_backward)
--- a/src/nativeops.jl
+++ b/src/nativeops.jl
@@ -51,8 +51,14 @@ end
 =#
 need_top_grad(:: Operator) = true
 
+#=doc
+.. function:: declare_backward_dependency(op :: Operator, out_grad, in_data, out_data)
+=#
+function declare_backward_dependency(:: Operator, out_grad, in_data, out_data)
+end
+
 ###
-# NativeOpInfo mirrors the struct in include/mxnet/c_api.h and consists of five function
+# NDArrayOpInfo mirrors the struct in include/mxnet/c_api.h and consists of five function
 # pointers that work as callbacks. Each p_... ia a opaque pointer that contains the
 # necessary information to call the right function.
 #
@@ -61,27 +67,31 @@ need_top_grad(:: Operator) = true
 #
 # Todo: Cleanup tasks.
 ###
-immutable NativeOpInfo
+immutable NDArrayOpInfo
   forward :: Ptr{Void}
   backward :: Ptr{Void}
   infer_shape :: Ptr{Void}
   list_outputs :: Ptr{Void}
   list_arguments :: Ptr{Void}
+  declare_backward_dependecy :: Ptr{Void}
+
   p_forward :: Ptr{Void}
   p_backward :: Ptr{Void}
   p_infer_shape :: Ptr{Void}
   p_list_outputs :: Ptr{Void}
   p_list_arguments :: Ptr{Void}
+  p_declare_backward_dependency :: Ptr{Void}
 
   function NativeOpInfo(op :: Operator)
     # infer_shape, list_args, list_outputs are called directly and use dynamic dispatch,
     # for finding the correct operator.
-    p_is, p_la, p_la = pointer_from_objref(op)
+    p_is, p_la, p_la, p_dbd = pointer_from_objref(op)
 
-    c_wrapper_fb = cfunction(_wrapper_fb, Void, (Cint, Ptr{Ptr{Cfloat}}, Ptr{Cint}, Ptr{Ptr{Cuint}}, Ptr{Cint}, Ptr{Void}))
+    c_wrapper_fb = cfunction(_wrapper_fb, Void, (Cint, Ptr{Ptr{Void}}, Ptr{Cint}, Ptr{Void}))
     c_wrapper_infer = cfunction(_wrapper_infer, Void, (Cint, Ptr{Cint}, Ptr{Ptr{Cuint}}, Ptr{Void}))
     c_wrapper_list_outputs = cfunction(_wrapper_list_outputs, Void, (Ptr{Ptr{Ptr{Cchar}}}, Ptr{Void}))
     c_wrapper_list_arguments = cfunction(_wrapper_list_arguments, Void, (Ptr{Ptr{Ptr{Cchar}}}, Ptr{Void}))
+    c_wrapper_declare_backward_dependency = cfunction(_wrapper_declare_backward_dependency, Cbool, Bool, (Ptr{Cint}, Ptr{Cint}, Ptr{Cint}, Ptr{Cint}, Ptr{Ptr{Cint}}, Ptr{Void}))
 
     # Setting up for handling backward/forward. Each function has a condition for a task
     # to wait on and a libuv callback that notifies that condition, to handle the call.
@@ -129,7 +139,8 @@ immutable NativeOpInfo
     end
 
     new(c_wrapper_fb, c_wrapper_fb, c_wrapper_infer, c_wrapper_list,
-        c_wrapper_list, p_f, p_b, p_is, p_lo, p_la)
+        c_wrapper_list,  c_wrapper_declare_backward_dependency,
+        p_f, p_b, p_is, p_lo, p_la, p_dpd)
   end
 end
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by pluskid** [comment]
You might want to use an actual object instead of mere type in the API. e.g. 
```julia
function list_output(op::Softmax)
  ...
end
```
As my impression is that an operator could actually have some properties, like `name`, etc.

vchuravy:
I want to formulate it threadsafe from the get go and I don't know how to do that (yet) -> https://groups.google.com/forum/#!topic/julia-dev/9J1GYfCyVpE

pluskid:
I dig a bit into the discussions on mailing list. Just to summarize based on my understanding:

1. For the kvstore callback (#18), since @tqchen mentioned, that callback is guaranteed to be called in the main thread, there is no issue of the current implementation. We will probably go with the current way since it is somehow more efficient for some unknown reason.
2. For the operator case, since the callback is executed from a different thread, and Julia GC is not thread safe (yet), the **only** safe thing to do in the callback is `ccall(:uv_async_send, ...)`. So we will have to make this a shallow wrapper that does nothing but notify an actual async worker in the main thread to do the job. This shallow callback can **not** do waiting as context switching is definitely going to cause issue. So we will need to modify `libmxnet` to add ability to get notified when the Julia async worker finished the actual computation of the operator.

vchuravy:
@pluskid Exactly. also relevant https://github.com/dmlc/MXNet.jl/pull/18#issuecomment-154132002

I am currently trying to implement async safe callbacks in OpenCL.jl with the ability to pass data along, but I am running into issues. I will keep you updated as it goes. https://github.com/JuliaGPU/OpenCL.jl/pull/87

tqchen:
Sorry to keep you wait for a while, this PR #667 enables asynchronize OP.

I would recommend make an update to NativeOp to enable async mode, that exposes a callback to C API, which does some finalization stuffs(e.g. copy data back to TBlob), and call async_on_compelete.

pluskid:
@tqchen Thanks for adding this! I had a closer look at the nativeop API. It's a bit strange to expose the user some raw pointer to operate with. I think it might be better to use `NDArray` directly since it has nice async execution scheduling. The user could choose to use `NDArray` operations to implement the operator (which is potentially faster and runs on GPU), or if he want to work with numpy array or Julia array, copying the data from and to an `NDArray` is super easy in the script side (we could also provide a simple default wrapper for this case). What do you think?

piiswrong:
@pluskid That was what I was going to do originally, but due to some reason I went this way instead.
You are welcome to refactor it.

vchuravy:
@pluskid @piiswrong I will look into the refactoring today, for both the python side and the julia side.
@tqchen Thanks for loonking into async ops.

piiswrong:
@vchuravy remember to set the ctx of the ndarraies to OpContext.ctx

pluskid:
@piiswrong Thanks! @vchuravy Great! I'll leave it to you now then.

tqchen:
There was a bit interfacing issue that we did not yet expose ndarray as operator.

- The callback Julia operations are meant to be ran by another worker thread, while itself execute in kinda of synchronized manner.
- NDArray operations are usually executed by main thread, and itself runs in async manner.

So if synchronize copy is introduced in the operator, the operator itself blocks in the memcpy call (of course this could also be ran by another worker thread), which is OK. Current operator was kinda of simplified to assume it actually does the heavy lifting works, copy to CPU and calls kinda of synchronize callback in native language.

To support NDArray tracking, we will need to introduce operator interface with NDArray, while most wrapper should still go with the synchronized real action.  This complicates a bit with c++11 issues, as c++11 support by nvcc was introduced, but seems not yet perfect, and current cuda related code are not yet c++11. But I am sure we will find a way.

For short term, making native julia and numpy op work would be a good first step

piiswrong:
@vchuravy BTW, you probably will be interested in https://github.com/dmlc/mxnet/pull/668
This is a runtime compile engine for CUDA. You can write kernels in python or julia

piiswrong:
@vchuravy when do you expect to finish C++ side refactor? I also want that feature, maybe we can setup an API standard and divide labor

vchuravy:
@piiswrong I was thinking of later tonight, but if you want go ahead and I focus on the Julia side.

vchuravy:
@pluskid I made some progress, but this is primarily focused on the conceptual side of how things *should* work. I am now working on getting it to a state where it actually works.

pluskid:
@vchuravy Thanks! With the end of the semester settled and coming back from conference, I think I will probably finally be able to resume a relatively stable commitment to the project soon. What is the current status and current blocking item? Let me know if you need any help on this side.

vchuravy:
I still have to write the entire points and the rtc interface. After that
it is mostly going to be weeding out bugs and making sure everything is
thread safe.

vchuravy:
@pluskid Conceptually everything is in place, but currently the following still seqfaults.

```julia
using MXNet
immutable Softmax <: mx.Native.Operator end
op = Softmax()
info = mx.Native.NDArrayOpInfo(op)
pstring = bytestring("0x", hex(reinterpret(UInt, pointer_from_objref(info))))
mx._Native(name = :test, info = pstring)
````

**Comment by vchuravy** [comment]
Initializing a NativeOperator now  works without seqfaulting.One major issue where I am unsure how to best handle that is the lifetime of Julia objects that we return via pointers to mxnet. One of those examples would be `list_arguments`. The Julia function returns an array of strings, that then gets converted to an array of pointers and we then store the pointer of that array to mxnet. When the GC kicks in the strings will get collected and the memory addresses will be void.

**Comment by vchuravy** [comment]
@pluskid 

I would like to define
```julia
function Base.call(op :: Operator; kwargs...)
  info = NDArrayOpInfo(op)
  pstring = bytestring("0x", hex(reinterpret(UInt, pointer_from_objref(info))))
  mx._Native(info = pstring, kwargs...)
end
```

So that the following would work

```julia
immutable SoftmaxOp <: mx.Native.Operator end; 
Softmax = SoftmaxOP(); 
Softmax(name = :Softmax)
```
But the function gets precompiled and it doesn't find `_Native`any ideas how circumvent that, short of using `@generated function`?

vchuravy:
The initialisation phase goes through correctly now and I added an example softmax function (that I still have to define correctly, but is enough as of now). 

I am currently encountering the problem that the tags (https://github.com/vchuravy/MXNet.jl/blob/b6a3f28dd4e2cd42b1dfb722dc3e7536feb6fa1e/src/nativeops.jl#L306) are nonsensical (probably corrupted or they went out of scope). Instead of getting 0 ... 4 I get values like  -235237891

src/nativeops.jl
```diff
@@ -0,0 +1,336 @@
...
+.. class:: Operator
+
+   Abstract type user have to derive from to implement Operators in Julia
+=#
+abstract Operator
+
+#=doc
+.. function:: forward(op :: Operator, in_data, out_data)
+=#
+function forward(:: Operator, in_data, out_data)
```
pluskid: Can you annotate the types of `in_data` and `out_data`, also for all the following functions?

src/nativeops.jl
```diff
@@ -0,0 +1,336 @@
...
+  out_data[0][:] = in_data[0]
```
pluskid: What is the type of `in_data` and `out_data`? What does indexing `0` mean here? (and followings)

src/nativeops.jl
```diff
@@ -0,0 +1,336 @@
...
+      # Get pointer to array
+      tshapes = unsafe_load(tensor_shapes, i)
+      # Copy values
+      for j in 1:jj
+        r_shapes[j] = unsafe_load(tshapes, j)
+      end
+      push!(shapes, r_shapes)
+    end
+
+    ishape, oshape = infer_shape(op, shapes)
```
pluskid: The shapes should be reversed before passing into `infer_shape` and then reversed before passing back to libmxnet. Because of row-majoring vs col-majoring.

src/nativeops.jl
```diff
@@ -0,0 +1,336 @@
...
+    end
+
+    new(c_wrapper_fb, c_wrapper_fb, c_wrapper_infer, c_wrapper_list_outputs,
+        c_wrapper_list_arguments,  c_wrapper_declare_backward_dependency,
+        p_f, p_b, p_is, p_lo, p_la, p_dbd)
+  end
+end
+
+###
+# Infer and list are called in sync.
```
pluskid: Why those callbacks could be called in sync? Could functions like `unsafe_load` possibly trigger gc?

pluskid:
@vchuravy I had some inline comments. I'm still in holiday-mode so please ignore me if I said something nonsensical.

For the memory issue in the functions like `list_arguments`. I think instead of asking the user to implement callback functions `list_arguments` to return a list of strings, which might be GC-ed any time, we could ask (via convention) that when the user define a new operator. The operator object should have a field called `arguments`, which contains a list of strings (similarly for `outputs`, etc.). So as long as the operator object itself is not GC-ed, those strings will not be GC-ed. Though I'm not very sure if there is a good way to automatically keep an operator from being GC-ed.

src/nativeops.jl
vchuravy: Copied that over from python by mistake. I decided instead of defining default function to throw a method error here.

vchuravy:
@pluskid I address most of your comments. I think we need a generic way of ensuring that the memory addresses we are passing to libmxnet stay valid, currently there are many places where values could get gc'd. Most likely a `WeakKeyDict` would do the job (except for the Operator itself...), I will look into that over the coming days. 

I have two questions. Do you know what the following means?
```
heimdallr:~/src/MXNet.jl/examples (vc/nativeops)> julia julia-softmax.jl 
INFO: Start training on [CPU0]
INFO: Initializing parameters...
ERROR: LoadError: ReadOnlyMemoryError()
 [inlined code] from /home/wallnuss/.julia/v0.4/MXNet/src/base.jl:53
 in __infer_shape#3__ at /home/wallnuss/.julia/v0.4/MXNet/src/symbolic-node.jl:253
 in init_model at /home/wallnuss/.julia/v0.4/MXNet/src/model.jl:98
 in _init_model at /home/wallnuss/.julia/v0.4/MXNet/src/model.jl:236
 in fit at /home/wallnuss/.julia/v0.4/MXNet/src/model.jl:334
while loading /home/wallnuss/src/MXNet.jl/examples/julia-softmax.jl, in expression starting on line 56
```

also before using 4b907348 I was getting the following. (This might be related to https://github.com/dmlc/mxnet/pull/672#issuecomment-158680454)
```
signal (11): Segmentation fault
pthread_mutex_lock at /usr/bin/../lib/libpthread.so.0 (unknown line)
unknown function (ip: 0x7f2a9f5400f5)
_ZNSt5mutex4lockEv at /home/wallnuss/src/mxnet/lib/libmxnet.so (unknown line)
_ZNSt10lock_guardISt5mutexEC1ERS0_ at /home/wallnuss/src/mxnet/lib/libmxnet.so (unknown line)
_ZN5mxnet6engine11ThreadedVar13ready_to_readEv at /home/wallnuss/src/mxnet/lib/libmxnet.so (unknown line)
_ZN5mxnet6engine14ThreadedEngine10WaitForVarEPNS0_3VarE at /home/wallnuss/src/mxnet/lib/libmxnet.so (unknown line)
_ZNK5mxnet7NDArray10WaitToReadEv at /home/wallnuss/src/mxnet/lib/libmxnet.so (unknown line)
MXNDArrayWaitToRead at /home/wallnuss/src/mxnet/lib/libmxnet.so (unknown line)
_wait_to_read at /home/wallnuss/.julia/v0.4/MXNet/src/base.jl:53
forward at /home/wallnuss/.julia/v0.4/MXNet/src/ndarray.jl:691
_entry_forward at /home/wallnuss/.julia/v0.4/MXNet/src/nativeops.jl:331
anonymous at /home/wallnuss/.julia/v0.4/MXNet/src/nativeops.jl:139
unknown function (ip: 0x7f2cb3ba6683)
unknown function (ip: (nil))
Segmentation fault (core dumped)
```

src/nativeops.jl
vchuravy: See https://github.com/dmlc/MXNet.jl/pull/18#issuecomment-154134731 only `forward` and `backward` are async.

mli:
Can one of the admins verify this patch?

vchuravy:
@mli what do you mean with verify? This is sadly still WIP and I didn't have time to work on this.

pluskid:
The `ReadOnlyMemoryError` is from Julia side and looks quite strange. Maybe the memory is already corrupted somewhere before reaching that place.

mli:
Ah, that is an auto reply of Jenkins. I'll figure out how to disable it later

vchuravy:
For anybody looking into this. I don't think that this will work safely till julia `v0.5. Then we will have experimental support for multi-threading and can support this properly.

pluskid:
thanks! is there a link to the mentioned multi thread support in julia v0.5?

vchuravy:
The original PR is https://github.com/JuliaLang/julia/pull/13410 and there has been more work on this since then. (see https://github.com/JuliaLang/julia/pull/15206)

The documentation lives at http://docs.julialang.org/en/latest/stdlib/parallel/#multi-threading

pluskid:
@vchuravy Thank you very much! I think at this point it makes sense to wait for Julia v0.5 before hacking this.

vchuravy:
So I think this is actually getting into the realm of the possible. The trick is to use put a lock around `:uv_async_send` and potentially making the entire call blocking.

Right now I am running into the issue that if I make the entire call blocking, the execution engine seems to stall and because operations on NDArray are blocking.

@piiswrong do I see it right that I should build this on top of the `Custom` operator and that `NativeOp` and `NDArrayOP` are deprecated? And am I also correct in assuming that https://github.com/dmlc/mxnet/blob/916dc9670abe2b64666c076bfff9f8feac23b8d6/src/operator/custom.cc#L75 is supposed to be blocking?

piiswrong:
@vchuravy Yes use custom op. The others are deprecated.

It's subtle. It can be either blocking or non blocking. However, if you do anything blocking in forward, you need to set the env variables such that there are more than 2 cpu/gpu worker threads

vchuravy:
Non blocking I would have to push a new sync point? Similar to how NVRTC is
implemented in python?

piiswrong:
No syncing is needed from the frontend side

PR #18, opened 2015-11-05
rewrite kvstore callback to be threadsafe

vchuravy:
Julia v0.4 is not threadsafe and we are not allowed to call any
functions that would allocate memory or invoke GC. (see https://groups.google.com/d/msg/julia-dev/9J1GYfCyVpE/-6KWFxD0AQAJ)

Even though it is not causing problems right now, it might be problematic in the future.

### DIFF
--- a/src/kvstore.jl
+++ b/src/kvstore.jl
@@ -107,9 +107,9 @@ end
 # extra handle parameter of the API to pass the updater object around. Fix this when someday
 # full closure cfunction is supported in Julia.
 function _kvstore_update_wrapper(index::Cint, nd_recv::MX_handle, nd_local::MX_handle, updater::Ptr{Void})
-  x = unsafe_pointer_to_objref(updater)
   updater_func = unsafe_pointer_to_objref(updater) :: Function
-  updater_func(Int(index), NDArray(MX_NDArrayHandle(nd_recv)), NDArray(MX_NDArrayHandle(nd_local)))
+  cb_packaged = Base.SingleAsyncWork(_ -> updater_func(Int(index), NDArray(MX_NDArrayHandle(nd_recv)), NDArray(MX_NDArrayHandle(nd_local))))
+  ccall(:uv_async_send, Void, (Ptr{Void},), cb_packaged.handle)
   return nothing
 end
 function set_updater(self :: KVStore, updater :: Function)


pluskid:
@vchuravy I tested this implementation a bit, it seems there is quite a bit overhead. Running the cifar10.jl, the speed drops from ~2200 samples/sec to ~1800 samples/sec.

If I understand correctly, do you mean if in the current implementation, the callback triggers some GC related functions (for example), then the program will fail or go into an undefined state?

Also I looked the post you did in the mailing list. It seems there is still not a good way to wait for the callback to run?

vchuravy:
I remember when I implemented this for OpenCL.jl that Julia (v0.3) would seqfault when the callback would be doing anything interesting, e.g. allocating memory, doing IO, calling library functions with side-effects. Certainly calling into the runtime from a different thread without using `uv_async_send` is undefined behaviour. The inner workings of uv_async are detailed in [1]

The second question is how to handle this with the current implementation of `NativeOP` in mxnet. Since the callback is executed asynchronously the function call from the mxnet side is returning before the callback is finished, so we would have to wait and I am not sure what the best/portable way of doing that is. (If Julia would be using libuv 1.X I would know how to do it.) Do you know how threading is implemented in mxnet?

I am wondering where the performance penalty is coming from. It might be that the additional anonymous function involved is hurting performance. For further reference the implementation of `SingleAsyncWork` is in [2].

[1] http://docs.libuv.org/en/v1.x/async.html
[2] https://github.com/JuliaLang/julia/blob/2d821670d2516cd38b51710b07b3eb18f191cd1b/base/stream.jl#L586

vchuravy:
Also of interest might be what `ZMQ` is doing here:
https://github.com/JuliaLang/ZMQ.jl/blob/master/src/ZMQ.jl#L316-L334

vchuravy:
I have an implementation on mind that might solve both problems at once. I
will work on that tomorrow (JST here)

tqchen:
I see what you mean, as the operations in the julia side was not thread safe, so an async invoke to the actual julia function is invoked. There are two solutions, one short term, another one longer term

Quick Solution
------------------
 Add wait in the async function call in the julia side.

Another Solution
---------------------
The fact is, mxnet's internal function call is also async, with tracked dependency, see [note engine](http://mxnet.readthedocs.org/en/latest/developer-guide/note_engine.html) and https://github.com/dmlc/mxnet/blob/master/include/mxnet/engine.h

The actual engine interface accepts a async function call like
```
Push(std::function<void(OnCompleteCallback) >, other dep related args)
```

So the true requirement is to ask the pushed function(in this case forward and backward, to call OnCompleteCallback when finished. 

***This means the function call can be asynced***, as long as OnCompleteCallback is called. Well this means calling a callback from Julia side when the action finishes and I don't know how easy that was, but might be interesting.

Current implementation of Operator assumes all the forward/backward are synced, so oncomplete was called right after the function get executed here https://github.com/dmlc/mxnet/blob/master/src/symbol/graph_executor.cc#L244  , maybe could be potentially modified

tqchen:
One more note, the callback for kvstore and for Operator was different

- KVStore callback was intended to be called during ***to issue commands*** to the engine, so in current implementation, it is invoked by the thread that calls the push and pull, i.e. same thread as the master thread in Julia
    - So I guess there won't be threading issue here
- The forward/backward are "the commands itself", and will be called asynchronously in another thread when dependency is fulfilled, which fits my last note.

vchuravy:
@tqchen in #16 (especially 00e24c6) I went for the quick solution.

@pluskid I could do what I did in 00e24c6 and try to improve the performance here, but if @tqchen doesn't think it will be a problem and we haven't run into problems yet we might be fine with the status quo.

Re:  as long as OnCompleteCallback is called. Well this means calling a callback from Julia side when the action finishes and I don't know how easy that was, but might be interesting.

That should be fairly straight forward.

tqchen:
I can confirm that the updater callback in kvstore currently is called from the same thread of the thread that calls the push/pull and all major C API calls

vchuravy:
@tqchen So only `forward` and `backward` are called from different threads? If so using `onCallback` would solve my problem #16 and simplify `infer_shape`, `list_outputs`, `list_arguments`.

Should I create an issue in the main repo, I don't yet understand MXNet well enough to make that change.

tqchen:
Yes, exactly. That will involved a refactor of current operator interface, which may need sometime to happen.
So I how about let us go with the sync solution you have for now, open an issue at main repo, keep it in mind and switch back to callback solution later?

vchuravy:
The sync solution uses illegal behaviour (https://groups.google.com/d/msg/julia-dev/9J1GYfCyVpE/xPUFFiktAgAJ) and won't work for the functions that are called by threads.

tqchen:
I see, I think we will run a refactor in one or two weeks, and will add the async operator to support this. Maybe we can wait until then

vchuravy:
That would be great. I wan't to switch from Mocha to MXNet, but I have research being depending on the ability to prototype things in Julia. So if you need any help figuring out how that interface should work. I would be more than willing to help out.

pluskid:
I guess instead of making all operator async, just make the native operator async or a different native async operator seems to be minimum changes.

PR #23, opened 2015-11-11
[RFC] reformultate accuracy with multi_output in mind

vchuravy:
Handling multi_output is a bit tricky because the inxed for the softmax output is inbetween the index for sample and the dimensions.

### DIFF
--- a/docs/api/io.rst
+++ b/docs/api/io.rst
@@ -317,7 +317,7 @@ libmxnet data providers
    
    
    :param prefetch_buffer: Backend Param: Number of prefetched parameters
-   :type prefetch_buffer: , optional, default=4
+   :type prefetch_buffer: long (non-negative), optional, default=4
    
    
    :param rand_crop: Augmentation Param: Whether to random crop on the image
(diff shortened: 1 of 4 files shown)


pluskid:
What is the behavior? Is it averaging or summing over the multiple output dimension? Could you document the behavior?

Also, some code from CPU backend of Mocha.jl could probably be borrowed. For example, see [this](https://github.com/pluskid/Mocha.jl/blob/master/src/layers/accuracy.jl#L54).

A side note: I was really bitten by `sub`, it was slow and sometimes leads to strange bugs. So I stopped using it. I heard there was a big rework on sub-arrays in Julia but have not checked the update yet.

vchuravy:
@pluskid when did you use sub, version v0.4 it is quite nice and certainly better than to create copies of the arrays, but I could do without and just inline indmax.

So the current behaviour is summing over all dimensions and then averaging, so in case of a pixel-wise prediction it calculates the mean accuracy per pixel, where would be the appropriate place to document this?

pluskid:
You can add to the comments [here](https://github.com/dmlc/MXNet.jl/blob/master/src/metric.jl#L42), and run `build-api.jl` in `docs` directory.

vchuravy:
@pluskid What would be the best way to check how much time is spend on calculating the accuracy?

Training times went from 4000s to 1100s in my case which seems a bit weird.

PR #25, opened 2015-11-12
Adds Mean Squared Error evaluation metric

Andy-P:


vchuravy:
This won't work for multi dimensional softmax output (see #23)

pluskid:
vchuravy this is for regression output. We do need to modify it for multi dimensional output though.

PR #28, opened 2015-11-16
Different Xaiver variants

vchuravy:
~Small fixup for Xavier~

This now implements a set of different variants. The one I don't know where it comes from is the original one (also used by the Python interface https://github.com/dmlc/mxnet/blob/bf988bf650daae4e7e6736221fa81e8443a45d5b/python/mxnet/initializer.py#L99). It was changed from the caffe_avg (3 /((in + out)/2) to 3 / (in +out) in https://github.com/dmlc/mxnet/commit/189fbc1b09d7f0d8ddbf6f41083911d04142a849 by @antinucleon )

If there isn't any good reason for having it I would suggest dropping it to minimize the options.

### DIFF
--- a/src/initializer.jl
+++ b/src/initializer.jl
@@ -104,14 +104,60 @@ end
 
    The initializer documented in the paper [Bengio and Glorot 2010]: *Understanding
    the difficulty of training deep feedforward neuralnetworks*.
-=#
-immutable XaiverInitializer <: AbstractInitializer
-end
 
-function _init_weight(self :: NormalInitializer, name :: Base.Symbol, array :: NDArray)
+   There are several different version of the XaiverInitializer used in the wild.
+   The general idea is that the variance of the initialization distribution is controlled
+   by the dimensionality of the input and output. As a distribution one can either choose
+   a normal distribution with μ = 0 and σ² or a uniform distribution from -σ to σ.
+
+   Several different ways of calculating the variance are given in the literature or are
+   used by various libraries.
+  
+   - original [Bengio and Glorot 2010]: σ² = 2 / (in + out)
+   - msra [K. He, X. Zhang, S. Ren, and J. Sun 2015]: σ² = 2 / in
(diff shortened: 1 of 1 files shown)


vchuravy:
So I was wondering currently Xavier is 

````
function _init_weight(self :: NormalInitializer, name :: Base.Symbol, array :: NDArray)
  dims    = size(array)
  fan_in  = prod(dims[2:end])
  fan_out = dims[1]
  scale   = sqrt(3 / (fan_in + fan_out))
  rand!(-scale, scale, array)
end
````

Following Bengio and Glorot it shouldn't it be:
````
  var   = 2 / (fan_in + fan_out) # Beignot 
  randn!(0.0, var, array)
````
or what caffee is using when Norm = Average
https://github.com/BVLC/caffe/blob/603cbfb97767d1b9ebf102200646f5df237d1749/include/caffe/filler.hpp#L127-L166

````
  scale   = sqrt(3 / ((fan_in + fan_out) / 2))
  rand!(-scale, scale, array)
end
````

Or what http://arxiv.org/abs/1502.01852 is suggesting

````
  var   = 2 / fan_in
  randn!(0.0, var, array)
````

vchuravy:
I was thinking of parameterizing XaiverInitializer by  the distribution and variant and then using multiple-dispatch, but that would noot have been backwards compatible.

vchuravy:
Also :MICROSOFT should be :MRSA and the whole thing needs more documentation

pluskid:
Thanks a lot! Looks good to me. Can you document it? Also I think if symbol is used, lower case names should be used to keep things consistent, and an error or assert should be made to ensure the parameters are valid candidates to avoid users having typo in the names and fallback to default behavior without noticing it, otherwise, can use @enum to define uppercase enum names explicitly.

vchuravy:
@pluskid  Let's hold onto releasing a new version till https://github.com/dmlc/mxnet/pull/610 has landed so that our API is equivalent.

PR #35, opened 2015-11-23
basic interface for setting and getting attributes

vchuravy:
In response to https://github.com/dmlc/mxnet/pull/685/

I suppose we would want as API:

````julia
mx.Data(name=:data, attributes = Dict(:key, "value"))
````

### DIFF
--- a/src/symbolic-node.jl
+++ b/src/symbolic-node.jl
@@ -123,6 +123,35 @@ function get_internals(self :: SymbolicNode)
 end
 
 #=doc
+.. function:: get_attr(self :: SymbolicNode, key :: Symbol)
+
+   Get attribute attached to this :class:`SymbolicNode` belonging to key.
+=#
+function get_attr(self :: SymbolicNode, key :: Symbol)
+  key_s = bytestring(string(key))
+  ref_out = Ref{Cstring}()
+  ref_success = Ref{Cint}(-1)
+  @mxcall(:MXSymbolGetAttr, (MX_handle, Cstring, Ref{Cstring}, Ref{Cint}), self, key_s, ref_out, ref_success)
+  if ref_success[] == 1
+    return bytestring(ref_out[])
+  else
+    throw(KeyError(key))
+  end
+end
+
+#=doc
+.. function:: set_attr(self:: SymbolicNode, key :: Symbol, value :: AbstractString)
+
+   Set the attribute key to value for this :class:`SymbolicNode`.
+=#
+function set_attr(self :: SymbolicNode, key :: Symbol, value :: AbstractString)
+  key_s = bytestring(string(key))
+  value_s = bytestring(value)
+
+  @mxcall(:MXSymbolSetAttr, (MX_handle, Cstring, Cstring), self, key_s, value_s)
+end
+
+#=doc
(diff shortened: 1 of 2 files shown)


pluskid:
Thanks! Yes, that kind of interface is better, and we should encourage the user to set the attributes at the beginning instead of modifying it dynamically.

pluskid:
About the interface for specifying the attributes upon creating the symbol (if you are working on that):

* I prefer `attrs` over `attributes`, just to be consistent with `get_attr`, `set_attr`
* We should probably enforce that all the values must be `AbstractString`s.

vchuravy:
Also the key values? These are currently set to be symbols, but that is
just personal preference.

pluskid:
Symbols for key is fine imo and is kind of Julia-ish

tbreloff:
My 2 cents... keys/values should be symbols if you can swing it. There's too much baggage that comes with using strings (ascii vs utf, etc) that is best avoided if possible.

pluskid:
@tbreloff I agree with you. But for values, because they are suppose to represent arbitrary values (e.g. numbers, tuples, etc.), I don't think it make sense to use symbols. The values should actually be of type `Any`, but this will make the backend implementation super complicated, so we just go with string currently. The corresponding party who uses specific attributes is responsible to decode the actual value from the string.

tbreloff:
So then Dict{Symbol,Any} is probably best. I think it adds too much complexity to require all inputs to be strings, especially if "they are suppose to represent arbitrary values (e.g. numbers, tuples, etc.)"

vchuravy:
@tbreloff I agree with you that for the user working with Any is the nicest, but the backend can only handle strings. In theory we could serialize and deserialize attributes.

pluskid:
I would like to keep the interface here rather light-weighted. String should be OK for commonly used values (symbols, strings, and probably integers, etc.), and this is how the underlying parameters for all operators are passed back into the libmxnet backend.

Attaching complicated objects is possible only if the user explicitly serialize it. I would like to make the interface in this case as difficult to use as possible as this should only be considered a temp workaround for something that should probably be implemented in another way. The reason is that complicated / opaque attributes attached to a symbolic node, when dumped to JSON string, will be impossible to decode from another frontend (e.g. Python or R).

vchuravy:
While thinking about how I would use this for setting lr for layers I realized that the return value of `get_attr` should probably be a `Nullable{String}` and not throw a error, since otherwise there is no way of checking whether or not an attribute is available.

tqchen:
Just a reminder about ```set_attr``` function and its possible consequences is discussed here https://github.com/dmlc/mxnet/blob/master/include/mxnet/c_api.h#L475

pluskid:
@tqchen Thanks! I added a warning to the doc of `set_attr`, but probably we should rename it to `_set_attr` to discourage it from being called explicitly. 

@vchuravy Yes I think that makes sense. Alternatively, we could use similar thing as the `get` interface (see [here](http://docs.julialang.org/en/release-0.4/stdlib/collections/?highlight=get#Base.get)) as in Julia base -- allow the user to supply a default value. I personally find Julia `Nullable` a bit verbose and inconvenient to manipulate, but I'm fine with either option.

PR #36, opened 2015-11-24
attr interface for SymbolicNodes

vchuravy:
Follow up on #35.

I went with the Nullable abstraction, because we don't have `haskey` and it encapsulates what is happening underneath. 

@pluskid Where would the right place to document this be?

### DIFF
--- a/src/symbolic-node.jl
+++ b/src/symbolic-node.jl
@@ -126,6 +126,7 @@ end
 .. function:: get_attr(self :: SymbolicNode, key :: Symbol)
 
    Get attribute attached to this :class:`SymbolicNode` belonging to key.
+   :return: The value belonging to key as a :class:`Nullable`.
 =#
 function get_attr(self :: SymbolicNode, key :: Symbol)
   key_s = bytestring(string(key))
@@ -133,9 +134,9 @@ function get_attr(self :: SymbolicNode, key :: Symbol)
   ref_success = Ref{Cint}(-1)
   @mxcall(:MXSymbolGetAttr, (MX_handle, Cstring, Ref{Cstring}, Ref{Cint}), self, key_s, ref_out, ref_success)
   if ref_success[] == 1
-    return bytestring(ref_out[])
+    return Nullable{ByteString}(bytestring(ref_out[]))
   else
-    throw(KeyError(key))
+    return Nullable{ByteString}()
   end
 end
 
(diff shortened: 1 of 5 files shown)


src/symbolic-node.jl
```diff
@@ -133,9 +134,9 @@ function get_attr(self :: SymbolicNode, key :: Symbol)
   ref_success = Ref{Cint}(-1)
   @mxcall(:MXSymbolGetAttr, (MX_handle, Cstring, Ref{Cstring}, Ref{Cint}), self, key_s, ref_out, ref_success)
   if ref_success[] == 1
-    return bytestring(ref_out[])
+    return Nullable(bytestring(ref_out[]))
```
pluskid: Can you use `Nullable{AbstractString}` explicitly?

src/symbolic-node.jl
```diff
@@ -133,9 +134,9 @@ function get_attr(self :: SymbolicNode, key :: Symbol)
...
   else
-    throw(KeyError(key))
+    return Nullable()
```
pluskid: Can you use `Nullable{AbstractString}` explicitly?

src/symbolic-node.jl
```diff
@@ -164,10 +165,14 @@ end
    Create a symbolic variable with the given name. This is typically used as a placeholder.
    For example, the data node, acting as the starting point of a network architecture.
 =#
-function Variable(name :: Union{Base.Symbol, AbstractString})
+function Variable(name :: Union{Base.Symbol, AbstractString}; attrs = Dict{Symbol, ASCIIString}())
```
pluskid: Why `ASCIIString` is used instead of `AbstractString`?

pluskid:
@vchuravy Thanks! I had some inline comments.

BTW: the right place to document what?

vchuravy:
The right place to address the attr keyword for SymbolicNodes and Variables.

pluskid:
The place for `Variable` doc should go [here](https://github.com/dmlc/MXNet.jl/blob/master/src/symbolic-node.jl#L162), and the doc for general operators could be appended [here](https://github.com/dmlc/MXNet.jl/blob/master/src/symbolic-node.jl#L492).

src/symbolic-node.jl
```diff
@@ -159,15 +160,19 @@ function set_attr(self :: SymbolicNode, key :: Symbol, value :: AbstractString)
...
 
 #=doc
-.. function:: Variable(name :: Union{Base.Symbol, AbstractString})
+.. function:: Variable(name :: Union{Symbol, AbstractString})
 
    Create a symbolic variable with the given name. This is typically used as a placeholder.
    For example, the data node, acting as the starting point of a network architecture.
 =#
-function Variable(name :: Union{Base.Symbol, AbstractString})
+function Variable(name :: Union{Symbol, AbstractString}; attrs = Dict{Symbol, AbstractString}())
```
pluskid: Julia types not being covariant bite me a lot. This interface might be somehow inconvenient to use, as if the user write
```
attrs = Dict(:foo => "bar", :bar => "baz")
```
they will get a `Dict{Symbol,ASCIIString}` (or `UTF8String` depending on whether they used unicode), yet that is not a subtype of `Dict{Symbol,AbstractString}`, and the call will fail with a signature mismatch. 

A workaround that I'm currently using is in the signature, use general `attrs::Dict = Dict()` type, and then in the function body, do a explicit call to
```
attrs = convert(Dict{Symbol,AbstractString}, attrs)
```
attempt to get the right behavior and fail gracefully with meaningful error message. Can you modify it to be of this way? Thanks!

src/symbolic-node.jl
```diff
@@ -521,6 +527,12 @@ function _define_atomic_symbol_creator(hdr :: MX_handle; gen_docs=false)
       if k == :name; continue; end
       if isa(v, SymbolicNode)
         symbol_kws[k] = v
+      elseif k == :attrs
+        if isa(v, Dict)
+          attrs = v
```
pluskid: Can you also add
```
v = convert(Dict{Symbol,AbstractString}, v)
```
here?

vchuravy:
@pluskid done and also see the new test to ensure that it is the correct behaviour.

pluskid:
Thanks a lot!

PR #37, opened 2015-11-26
adds LearningRate.Inv from Mocha

vchuravy:
Adds the `Inv` LearningRate scheduler from Mocha


vchuravy:
@pluskid One question would be what the default values for power and gamma should be.

pluskid:
@vchuravy Power 0.5 is a good default choice, which gives around `1/sqrt{T}` convergence rate for subgradient descent of convex function. I don't have good suggestion for gamma. It will depend on the problem. But `1` or `0.9` or `0.1` might all be good.

PR #42, opened 2015-11-27
Fixes usuage of floor

vchuravy:
https://github.com/dmlc/mxnet/pull/717 a PR upstream added `floor`/`ceil`/`round,` which are now imported as operators, causing problems with the native Julia functions. Luckily there is only one usage.

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -41,7 +41,7 @@ end
 """
 function _split_inputs(batch_size :: Int, n_split :: Int)
   @assert(batch_size >= n_split)
-  per_split = floor(Int, batch_size / n_split)
+  per_split = Base.floor(Int, batch_size / n_split)
   counts    = Base.zeros(Int, n_split)+per_split
   extra     = batch_size - sum(counts)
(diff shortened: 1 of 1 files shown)


pluskid:
Thanks! We should actually import base.floor somewhere (in symbolic node I guess) then Julia will be able to do auto dispatch according to argument type.

vchuravy:
Hm, I think we should follow what we do for zeros and similar functions. Require user to be explicit and also be explicit in MXNet.jl.

pluskid:
Actually, it is only for `zeros` and `ones` because it is impossible to resolve the ambiguity. For other functions like `size`, `length` and even operators `+`, etc. You can do either `mx.size(ndarray)` or `size(ndarray)`. Both works fine.

maximsch2:
MXNet.jl one gets from `Pkg.add` is broken now due to this issue. Is it possible to tag a new version?

pluskid:
@maximsch2 Sure, let me do it now.

PR #76, opened 2016-04-11
add ccSGD

vchuravy:
Implements ccSGD following the python version.

Requires: https://github.com/dmlc/mxnet/pull/1820


vchuravy:
@pluskid Something is still odd, the accuracy is a lot lower then with the Julia SGD.

pluskid:
Thanks! Let us know if you find the issue causing the difference!

PR #79, opened 2016-04-14
Refactor multi-dimensional accuracy measure and add average cross-entropy

vchuravy:
I was reading the code in metric.jl again and I couldn't understand my own implementation of Accuracy anymore. So I rewrote it to be more obvious to what it does, but it is slightly less general.

Additionally this adds the average cross-entropy measure or logloss measure for multidimensional softmax.


pluskid:
Thanks! I'm thinking of disabling travis CI for Julia nightly during the period when main contributors do not have much time to keep up with the latest changes.

PR #82, opened 2016-04-18
Use openblas from Julia, instead of installing yet another BLAS library

staticfloat:
This PR removes the Linux package-managing stuff requiring users to install `libcblas`.  We know Julia already comes with `libopenblas`, and as there is no real reason I can think of to use `libcblas` over `libopenblas`, I suggest this change as it makes things easier and doesn't cause potential conflicts with systems that already have a BLAS library installed.

What do you think?

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -30,6 +30,11 @@ if !libmxnet_detected
 
   openblas_path = Libdl.dlpath(Libdl.dlopen(Base.libblas_name))
 
+  ilp64 = ""
+  if Base.blas_vendor() == :openblas64
+    ilp64 = "-DINTERFACE64"
+  end
+
   #--------------------------------------------------------------------------------
   # Build libmxnet
   mxnet = library_dependency("mxnet", aliases=["libmxnet", "libmxnet.so"])
@@ -51,6 +56,7 @@ if !libmxnet_detected
           `cp make/config.mk config.mk`
           @osx_only `cp make/osx.mk config.mk`
           `sed -i -s 's/USE_OPENCV = 1/USE_OPENCV = 0/' config.mk`
+          `sed -i -s "s/MSHADOW_CFLAGS = \(.*\)/MSHADOW_CFLAGS = \1 $ilp64/" mshadow/make/mshadow.mk`
           `cp ../../cblas.h include/cblas.h`
           `make USE_BLAS=openblas MSHADOW_LDFLAGS="$openblas_path" -j`
           `cp lib/libmxnet.so $_libdir`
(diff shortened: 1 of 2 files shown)


pluskid:
@staticfloat Yes this is a very reasonable argument. We actually thought about this. But the openblas shipped with Julia does not contain a header file. So the compilation of libmxnet net will fail. I'm wondering how did you resolve this issue?

staticfloat:
Ah, good call.  I didn't realize that the `cblas.h` header wasn't included already.  I've included a simplified version of the header from Netlib's distribution, which works on my machine, but of course this should be tested on multiple machines before being merged, as it's common for me to have dependencies laying around on my development machines that interfere with the build process and make things work when they shouldn't.  :)

staticfloat:
Alright, with this latest push, I have code that works on both my Mac Mini and by Linux box.  It's unfortunate that `libmxnet` installs itself as `libmxnet.so` no matter the underlying platform, but oh well.  That's not a big deal.  :P

pluskid:
@staticfloat Thanks for the contribution. I tend to think we need to be careful here. As I am not very sure if the same cblas.h could be used everywhere or if the cblas.h from Netlib could be used directly for openblas. Do you have any reference on whether there is a standard for the header file here?

staticfloat:
Yes, this is the BLAS header file distributed with the Netlib BLAS implementation, which is the reference standard.  It's linked to directly from the [Netlib website](http://www.netlib.org/blas/cblas.h), and is portable across platforms.  Because CBLAS is a standard, OpenBLAS should always be conformant to that standard, however, there is one thing that can change depending on how OpenBLAS is compiled; the definition of `CBLAS_INDEX`.  Right now, it's set to `size_t`, but if the user has an ILP64-enabled OpenBLAS (which is the default in from-source builds of Julia) then this header needs some modification.  I will work this up, and then we can test it on a wide variety of systems to ensure that it's functioning properly.

pluskid:
@staticfloat OK. Thanks for the efforts! I can test it on Mac OS, Ubuntu and CentOS. It would be great if anybody could test it on some windows machine.

staticfloat:
I've made some changes, cutting down the included `cblas.h` file to the minimum needed by us, and conforming it more to what OpenBLAS uses by using the OpenBLAS `cblas.h` file as a template.  I've also tested this on both ILP64 and non-ILP64 Julias on both Ubuntu and OSX, and it works on all four configurations, on 0.4 and 0.5.  I'm just waiting for Travis to get around to running it on the cloud, but I think this is ready for you to try on your computers now.

> It would be great if anybody could test it on some windows machine.

Since we don't build automatically on Windows, I'm not sure how they would test it, other than compiling manually, which would ignore all the changes I've made anyway.

staticfloat:
Looks like Travis tests on `0.4` pass now.

pluskid:
@staticfloat Great work! Thanks a lot!

PR #87, opened 2016-04-26
[RFC] Freeze layers

### LINKED ISSUE #63: freeze layers weight
Is there a way to freeze layer weights?

vchuravy:
Setting `attrs = Dict(:grad => "freeze")` on a layer causes it to be ignored in the when updating the weights.

We probably can make this a bit more efficient by setting grad_req=GRAD_NOP in bind for these layers.

This depends on https://github.com/dmlc/mxnet/pull/1965 and fixes #63.

### DIFF
--- a/src/executor.jl
+++ b/src/executor.jl
@@ -128,22 +128,31 @@ function bind(self :: SymbolicNode; kwargs...)
   bind(self, context, args; kwargs...)
 end
 
-function simple_bind(self :: SymbolicNode, ctx :: Context; grad_req :: GRAD_REQ=GRAD_WRITE, kwargs...)
+function simple_bind(self :: SymbolicNode, ctx :: Context;
+                     grad_req :: Union{GRAD_REQ, Dict{Symbol, GRAD_REQ}}=GRAD_WRITE,
+                     kwargs...)
   arg_shapes, out_shapes, aux_shapes = infer_shape(self; kwargs...)
   @assert(!isa(arg_shapes, Void), "Information not enough to perform complete shape inference")
 
   arg_arrays = NDArray[zeros(shape, ctx) for shape in arg_shapes]
   arg_names  = list_arguments(self)
-  if grad_req == GRAD_NOP
-    grad_arrays = Dict{Base.Symbol,NDArray}()
-  else
+
+  grad_arrays = Dict{Symbol,NDArray}()
+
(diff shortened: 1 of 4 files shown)


vchuravy:
Okay this doesn't work yet.

For testing I set in lenet
```julia
conv2 = @mx.chain mx.Convolution(data=conv1, kernel=(5,5), num_filter=50, attrs=Dict(:grad => "freeze"))
```

and after training load back the weights and compare them.

vchuravy:
The updater is never run on the frozen weights/biases but they still change over time. Is there anyplace else where they are adapted?

src/executor.jl
```diff
@@ -128,22 +128,31 @@ function bind(self :: SymbolicNode; kwargs...)
...
-      # if not in provided data, should be parameters
-      if !in(name, provided_data_names)
-        grad_arrays[name] = zeros(shape, ctx)
-      end
+    shapes = filter(x -> !in(x[1], provided_data_names), shapes)
+
+    # Remove all gradients for nop params
+    # if isa(grad_req, Dict{Symbol, GRAD_REQ})
+    #  shapes = filter(x -> grad_req[x[1]] != GRAD_NOP,shapes)
+    # end
```
vchuravy: Uncommenting the lines above leads to problems with how `fit` operates over grad_arrays. Thus we have to currently create them, but this should be fixed in the future to minimise memory consumption.

vchuravy:
With 9880bdf only weight_decay still happens..., but no adaptation anymore.

If I set weight_decay to 0 the weights stay constant. But it seems that ce599a9 is not enough.

vchuravy:
68180f7 should do the trick, but I don't have time to thoroughly test now.

pluskid:
@vchuravy Thanks a lot! There was recently a refactor in the Python side (https://github.com/dmlc/mxnet/pull/1854) that allows people to specify per-parameter learning rate. If you have time to take a look, that might help in this case. Definitely not super efficient, but setting the learning rate to 0 should be a very convenient way to temporarily disable the learning of a layer.

vchuravy:
@pluskid Has that been merged? The benefit this approach has is that MXNet won't compute the gradients and thus backwards can be faster than in other cases.

pluskid:
That has been merged. But I agree with you, this makes computation faster.

Issue #89, opened 2016-05-08
unzip is not available on windows


Mike43110:
Running the test suite, `io.jl` tries to spawn `unzip` which fails as unzip is not available.
Changing the command to `7z x $minst_dir` works however, while 7z is available in the Julia path, Julia may not necessarily be in the path. Possibly give a warning that unzip is not available on Windows?

PR #90, opened 2016-05-08
Added 7z alternative for mnist allowing test to run on windows

Mike43110:
Added try catch around the unzip and added `7z` extraction alternative.

If Julia is added to the path, `7z` will be present in the path. `unzip` is not present on windows which caused the `IO::MNIST` to fail.

Added try/catch around the run expressions with an error message about no extraction utility on path.


pluskid:
Thanks! This fixed #89.

Mike43110:
Just pushing my minor changes trying to get all the tests to pass! I don't think there are many windows users so I will try and keep the tests going as far as I can! (#91 forced me to read a lot of the source after all)

tkelman:
Note that `pipe` is deprecated in favor of `pipeline`

PR #91, opened 2016-05-08
Change where set_attr is called in _define_atomic_symbol_creator

Mike43110:
Where `set_attr` was originally called in
`_define_atomic_symbol_creator` would cause `list_attr` to return a
dictionary with extra bias and weight fields as seen here: `
Dict{Symbol,ByteString}(:convolution0_a=>"a",:convolution0_π=>"π",:convolution0_bias_a=>"a",:convolution0_weight_π=>"π",:convolution0_bias_π=>"π",:convolution0_weight_a=>"a",:data2_test=>"hallo!")`

The return now is:`Dict{Symbol,ByteString}(:convolution0_a=>"a",:convolution0_π=>"π",:data2_test=>"hallo!")`

The test still fails however as `list_attr` still contains the name of
the source before the key.

Unsure whether the tests need to be changed to reflect the current return symbols or whether `list_attr` should be modified to try and strip the name information from the keys.

### DIFF
--- a/src/symbolic-node.jl
+++ b/src/symbolic-node.jl
@@ -591,15 +591,15 @@ function _define_atomic_symbol_creator(hdr :: MX_handle; gen_docs=false)
     hint = lowercase($func_name_s)
     name = get!(DEFAULT_NAME_MANAGER, name, hint)
 
-    # set attrs
-    for (k, v) in attrs
-      set_attr(node, k, v)
-    end
-
     if length(args) != 0
       _compose!(node, name, args...)
     else
       _compose!(node; name=name, symbol_kws...)
+    end
+
+    # set attrs
+    for (k, v) in attrs
+      set_attr(node, k, v)
     end
 
     return node


pluskid:
Thanks! This is a nice catch. I think we need to add an option to the upstream `list_attr` to indicate whether to list recursively or only shallowly -- in the latter case, avoid prepending the symbol names.

vchuravy:
I think https://github.com/dmlc/MXNet.jl/blob/2b0d825ffa57d15498bf9e8c6e65b6c30011c7e9/src/model.jl#L372-L381 currently depends on this behaviour.

Mike43110:
I have made the changes in mxnet to add the shallow `listattr`. Also made the changes to the julia and python bindings as necessary.

Is the below behaviour the one you are addressing? Or is it just in reference to making a shallow `list_attr` upstreap?
`Dict{Symbol,ByteString}(:convolution0_a=>"a",:convolution0_π=>"π",:convolution0_bias_a=>"a",:convolution0_weight_π=>"π",:convolution0_bias_π=>"π",:convolution0_weight_a=>"a",:data2_test=>"hallo!")`

pluskid:
I have made a PR in the upstream dmlc/mxnet#2072 to add an interface to do a shallow list of attr (and the old recursive one). We should revert change here back as freezing feature and potentially some other features related to the optimizer need this. The existing unit test could be modified to call the shallow version of list attr.

pluskid:
@Mike43110 Now the issue with the attributes should be addressed! Please update to the latest MXNet.jl and libmxnet to test!

Mike43110:
Thanks!
Forgot to pull the new mxnet changes twice... 

All tests passing on windows!

PR #92, opened 2016-05-09
add support for ListAttrShallow

vchuravy:
Depends on https://github.com/dmlc/mxnet/pull/2072 in reference to https://github.com/dmlc/MXNet.jl/pull/91

Should `list_attr` be the shallow version?


pluskid:
Yes, I think it makes sense to let `list_attr` to be the shallow version. I will try to refactor the Python side later on according to suggestion in dmlc/mxnet#2070, and distinguish `list_attr` and `list_all_attr`. If people don't like changing existing function call, we can discuss about it at that point.

PR #94, opened 2016-05-11
Deduplicate and generalise metric update!

vchuravy:
MXNet allows for the design of networks that use the same label for multiple outputs. Instead of failing for these kinds of networks, warn the user and try to proceed.

The correct fix would be to keep track of which label belongs to which output, but currently we lose that kind of mapping.
@pluskid I was wondering after looking through `fit`, would you be open to a refactor of MXNet.jl to use `Dict{Symbol, NDArray}` instead of `Vector{NDArray}`?

This would also be helpful in cases like https://github.com/dmlc/MXNet.jl/blob/9880bdf57b350b2602684446d3bfe0f3a2a0094b/src/executor.jl#L149-L152 , where we need to create `NDArray` for non-existing gradients to not confuse the indexing in `fit`.

### DIFF
--- a/src/metric.jl
+++ b/src/metric.jl
@@ -36,6 +36,19 @@ set.
 =#
 abstract AbstractEvalMetric
 
+# Generic update! version
+function update!{T <: AbstractEvalMetric}(metric :: T, labels :: Vector{NDArray}, preds :: Vector{NDArray})
+  if length(labels) != length(preds)
+    Base.warn_once(
+      "The number of labels ($(length(labels))) does not correspond to the\
+      number of outputs ($(length(preds))). The calculated metric might not be accuracte.")
+  end
+  for (label, pred) in zip(labels, preds)
+    _update_single_output(metric, label, pred)
+  end
+end
+
+
 #=doc
 .. class:: Accuracy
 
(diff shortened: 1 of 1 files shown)


vchuravy:
Hm no travis tests anymore?

pluskid:
Travis is running. What do you mean?

@vchuravy I understand your point. I think the design here is that the user should write his own eval metric for a highly customized network. The number of outputs might well be different from the number of labels. For example, that happens when we implement RNNs using back-propagate-through-time, because the forward states are also included in the outputs, so it is more than the number of labels. So in summary:

- The standard case: same number of labels and outputs could be handled smoothly
- Slightly non-standard case: more outputs than labels, but the first K outputs corresponds to the labels in a correct order. It is good if the user could use an option to indicate that this is the expected behavior, thus turning off the warning.
- Most general case: complicated dependency between labels and outputs. The user should write his own eval metric. I agree it is good if you could come up with a helper base case (with dictionary mapping or something) to help the user to easily implement this general case.

vchuravy:
This implements 1-2, except for the user having to disable the waring.

The only problem I see is that `warn_once` is unexported from Base.

pluskid:
What I mean is that we might need an option to disable that `warn_once` if the author know what he is doing. Something like [this](https://github.com/dmlc/mxnet/blob/master/python/mxnet/metric.py#L334). But this looks good for now.

PR #97, opened 2016-05-24
[WIP/RFC]Type support for NDArray

vchuravy:
MXNet and mshadow support a range of different types. The Julia backend currently only works with Float32 and so I am hoping to extend it to support all types supported in the backend.

@pluskid Should we convert `NDArray` to `NDArray{T}` so that we get typestable methods? e.g. `getindex`...

### DIFF
--- a/src/ndarray.jl
+++ b/src/ndarray.jl
@@ -6,6 +6,7 @@ NDArray API
 # All the types supported by mshadow.
 typealias DType Union{Float32, Float64, Float16, UInt8, Int32}
 @enum TypeFlag kFloat32 kFloat64 kFloat16 kUint8 kInt32
+typealias DEFAULT_DTYPE Float32
 
 function toTypeFlag{T <: DType}(:: Type{T})
   if T == Float32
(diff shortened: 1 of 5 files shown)


pluskid:
@vchuravy using `NDArray{T}` might be better, but I think there might be some trade-off

- Converting `NDArray` to `NDArray{T}` might involve quite a large amount of work and refactoring
- Some interface might become unnecessarily complicated. Based on my understanding (might be incorrect), it seems that MXNet support hybrid data types in computing. If this is the case, then in the Julia side, if a function takes 3 `NDArray` parameters, we will have to define it as `function func{T1,T2,T3}(...)`.

vchuravy:
The second point is my main point of struggle right now, but not for `func(NDArray, NDArray)`, but for `func(Vector{NDArray})` where Vector could be `[NDArray{Float32}, NDArray{Float64}]` or `[NDArray{Float32}, NDArray{Float32}]`, the first one will be of type `Vector{NDarray{T <: Union{Float32, Float64, ...}}}` and the second one will be `Vector{NDArray{Float32}}`. Each of these version needs a different type constraint. 

As an example:
```julia
ERROR: LoadError: LoadError: MethodError: `_get_ndarray_inputs` has no method matching _get_ndarray_inputs(::ASCIIString, ::Array{MXNet.mx.NDArray{Float32},1}, ::Array{Symbol,1}, ::Bool)
Closest candidates are:
  _get_ndarray_inputs(::AbstractString, ::Array{MXNet.mx.NDArray{T<:Union{Float16,Float32,Float64,Int32,UInt8}},1}, ::Array{Symbol,1}, ::Bool)
  _get_ndarray_inputs(::AbstractString, ::Dict{Symbol,MXNet.mx.NDArray{T<:Union{Float16,Float32,Float64,Int32,UInt8}}}, ::Array{Symbol,1}, ::Bool)
```

The easiest thing would be to drop type annotations in those cases....

pluskid:
OK, I see. I think the main use of type annotation in Julia is to generalize specialized code so that computation is faster. But since in our case, all the computation is carried out in the C++ backend, it does not really matter. Therefore, using a general `NDArray` might be simple and good enough.

vchuravy:
That seems sensible. See 4ecc8a7 forhow it might end up in the end. Lots more to do, but this seems the least invasive.

vchuravy:
So this depends on https://github.com/dmlc/mxnet/issues/2260 for the `bind` tests. Following that model, prediction, fit setup is missing. The Julia arrays need to be of the correct type.

pluskid:
Great! I am fine with either of the two ways

- deprecate default type and force to always call the explicit typed version after several releases
- keep the default type but make it globally configurable -- might be a bit tricky to implement if we need to support Julia pre-compilation.

vchuravy:
We don't necessarily need to deprecate untyped zeros and ones. Julia already has a untyped zeros that defaults to Float64.

Otherwise I think this is ready to merge and I will add the high-level support later one. But I am working right now on extending mxnet proper to have better support.

pluskid:
@vchuravy Do you have idea why the travis is failing?

vchuravy:
`v0.5` needs changes to the buildscript. And the failure in `v0.4` is because `Float16` division might have a different error bounds in MXNet than in Julia.

vchuravy:
That was fast ;) I will make the necessary changes and submit a PR ASAP to make the test stable.

PR #98, opened 2016-05-27
Fix deprecation for v0.5 and adapt macro to expr changes

vchuravy:


REQUIRE
```diff
@@ -1,4 +1,5 @@
-julia 0.4
+julia 0.4+
```
tkelman: why did you add the `+`? that makes the package not installable on 0.4

vchuravy: I think in the first iteration I was planning on removing support vor `v0.4` completely before settling on `Compat`.

PR #115, opened 2016-08-10
Get MXNet.jl ready for v0.5

vchuravy:
Since I last updated the Julia bindings in preparation for the release v0.5 there where a bunch of changes with regards of handling `Cstring`. This fixes almost all deprecation while keeping support for v0.4.

At some point we should go through all `unsafe_wrap` and `unsafe_string` methods to establish who is responsible to free the memory associated with the pointer. Currently the assumption is in most cases that these pointers are going to be freed-ed by MXNet proper. 

Once we switch to only support v0.5 9e9c490 should be reverted and thus removing the last warnings.

Lastly this also contains some cleanup with regards to formatting of documentations, but we were also overwriting docs accidentally and I need to find a way to do this properly (v0.5 only).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
   - osx
 julia:
   - 0.4
+  - 0.5
   - nightly
 
 # dependent apt packages
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,3 +1,4 @@
+using Compat
 ################################################################################
 # First try to detect and load existing libmxnet
 ################################################################################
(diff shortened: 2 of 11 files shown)


pluskid:
Great! Thanks! It seems Julia v0.5 should be officially released soon. I'm thinking about refactoring some of the high-level APIs (at least port the Module API) to Julia once nnvm is finished (dmlc/mxnet#2944).

vchuravy:
So this should be good to merge. I will add a PR where I remove the deprecations in the examples.

What is NNVM? I only found a reference to Neuronal-Network-Visual-Model. An other thing we need to refactor is for our data-providers to also provide information about the dtypes of the inputs.

pluskid:
Thanks! NNVM is basically a replacement of the underlying symbolic graph execution engine. Most of the external APIs should not change much so I guess we do not need to worry too much. But one improvement might be that on-the-fly defining a new graph and run should be faster. It is still under internal development, maybe should come out in a few weeks.

PR #116, opened 2016-08-18
Cleanup `` and :class:

vchuravy:
This removes remnant `:class` and `` throughout the docs. We will need to remove a few more RST stuff, but as an example the `note` blocks can be replaced by `!!! note` and we could have something similar for warnings.

We should also take a look at https://github.com/JuliaDocs/DocStringExtensions.jl


pluskid:
It seems that the docs are not automatically updated recently. I will do some check later this week.  But do you happen to know what we might be mis-configuring or so?

PR #118, opened 2016-08-22
reorder docs for io.jl and bring them into the markdown age

vchuravy:
This improves the doc situation for `io.jl`. There might be a problem with `@autodocs` and `@docs` that needs to be solved in Documenter (I am still investigating).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ script:
 
 after_success:
   - source $TRAVIS/run_coverage.sh
+  - echo $TRAVIS_JULIA_VERSION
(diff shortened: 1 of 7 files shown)


vchuravy:
@pluskid I have seen documentation build failures on the v"0.6" of Julia today so that might be why they didn't autoupdate. (Per default the doc-update runs on nightly).

also x-ref: https://github.com/JuliaDocs/Documenter.jl/pull/125#issuecomment-241537833

vchuravy:
@pluskid The problem was not as I suspected v0.6, but rather attaching docs to `(self::SymbolicNode)(...)`

PR #123, opened 2016-08-24
handle kwargs for ndarray functions

### LINKED ISSUE #120: mx.transpose arbitrary axis?
mx.transpose seems only be able to transpose all the axis, how can I swap arbitrary axis of a symbolic NDArray? 

```
methods( mx.transpose )
transpose(in1::MXNet.mx.NDArray)
transpose(in1::MXNet.mx.NDArray, out1::MXNet.mx.NDArray) at ~.julia/v0.4/MXNet/src/ndarray.jl:1046
```

the python frontend seems have this functionality.

vchuravy:
this fixes #120 

This is similar to how python handles things, but it doesn't allow us to handle miss-typed kwargs gracefully.

@pluskid For transpose we might need to do some more work, since the expected order and axis's ids are in python order and not Julia order (thus backwards...)

### DIFF
--- a/src/ndarray.jl
+++ b/src/ndarray.jl
@@ -1035,6 +1035,15 @@ function _get_function_expressions(handle :: MX_handle, name)
   if name == :dot
     _use_vars.args[2:end] = flipdim(_use_vars.args[2:end], 1)
   end
+
+  if name == :transpose
+    transform = quote
+      kwargs = Any[key != :axes ? (key, arg) : (key, reverse(map(i->length(arg)-i, arg))) for (key, arg) in kwargs]
+    end
+  else
+    transform = :()
+  end
+
   stmt_call = quote
     local handle = _get_function($(QuoteNode(name)))
     _invoke_mxfunction(handle, $_use_vars, $_scalars, $_mut_vars; kwargs...)
@@ -1047,6 +1056,7 @@ function _get_function_expressions(handle :: MX_handle, name)
 
   func_def = quote
     function $name($(args...); kwargs...)
+      $transform
       $stmt_call
       $stmt_ret
     end
--- a/src/ndarray.jl
+++ b/src/ndarray.jl
@@ -885,17 +885,17 @@ with corresponding support (see `load`).
 * `filename::String`: path to the binary file to write to.
 * `data`: data to save to file. Data can be a`NDArray`, a `Vector{NDArray}`, or a `Dict{Base.Symbol, NDArray}`.
 """
-function save(filename::AbstractString, data::NDArray)
+function save(filename::String, data::NDArray)
   save(filename, [data])
 end
-function save(filename::AbstractString, data::Vector{NDArray})
+function save(filename::String, data::Vector{NDArray})
   @mxcall(:MXNDArraySave, (char_p, MX_uint, Ptr{MX_handle}, char_pp),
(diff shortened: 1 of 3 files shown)


pluskid:
Yes, this is tricky. I think we need to use some hooks to properly handle some special functions.

vchuravy:
It's especially tricky because we not only need to flip the order, we also need to change the indices (at least for the transpose case)?
`transform(axis) = reverse(map(i->length(axis)-i, axis))` Or is my intuition about this wrong?

vchuravy:
@piiswrong Can you chime in? I don't quite understand what axes transpose expects and you have added it in #2297

vchuravy:
@pluskid tests are failing because  webdocs.cs.ualberta.ca is unreachable.

Also there is a transpose for `SymbolicNodes` with an `axes` parameter. Should we also special case that?

pluskid:
Yes, we should special that, too.

I can still access the mnist.zip URL. Just restarted Travis CI to see if it get through this time. Meanwhile we can try to replace that URL with a more standard one.

src/ndarray.jl
```diff
@@ -1033,31 +1035,53 @@ function _get_function_expressions(handle :: MX_handle, name)
   if name == :dot
     _use_vars.args[2:end] = flipdim(_use_vars.args[2:end], 1)
   end
+
+  if name == :transpose
+    transform = quote
+      new_kwargs = Any[]
+      for (key, arg) in kwargs
```
pluskid: This is minor, but maybe using a list comprehension expression instead of for-loop + `push!` might be more efficient?

Could you also fix it for the symbolic version?

test/unittest/ndarray.jl
```diff
@@ -298,6 +298,15 @@ function test_eltype()
   end
 end
 
+function test_kwargs()
+  info("NDArray::kwargs")
+  dims1 = (2,3,4)
+
+  x = mx.empty(dims1)
+  tx = mx.transpose(x, axes=(2,1,3))
+  @test size(tx) == (3,2,4)
```
pluskid: Could you actually generate a random matrix, and use Julia's build-in swap axis function to compare against the array contents of the libmxnet transposed version?

pluskid:
@vchuravy I have updated the dataset URL, and the other PR has passed Travis CI testing. This PR is now in conflicts, could you update this one? Thanks!

vchuravy:
Rebased and updated.

We are special-casing `dot` for `NDArray`, but not for `SymboliNodes` is that intended?

pluskid:
@vchuravy No that was not intended. I think that was because many of those functions were only available for `NDArray` but not for symbolic back in the days.

pluskid:
@vchuravy I added special fix of dot for `SymbolicNode`. Also the handling of transpose has some small issue [here](https://github.com/dmlc/MXNet.jl/commit/f06d9943f51fdb77b8c2841aa70258eb4dbf7f30#diff-833e285b1e87d981a2f2dd5e21d257f6L633), as the `$func_name` is a `Function` instead of a `Symbol` in that context, so that `if` is never true.

The current way is however not very elegant, because those tests are called for every symbolic node function. A better way is to move those conditions outside of the function definition, like what we do for the `NDArray` functions. But I will leave it for now.

PR #124, opened 2016-08-24
add MultiMetric and reorder metric docs and update ACE and MultiACE for 1d labels

vchuravy:
This adds a simple metric that can be used to combine multiple metrics to see MSE and logloss at once.

@pluskid Ideas on how to test this?


pluskid:
I don't think we have unit tests for metrics. I am not sure what is the best way to test it, as the metrics are written in Julia, and there is no point to implement it again in the test case and compare against each other. Maybe testing some corner case would be fine (e.g. all incorrect leads to 0 accuracy, etc.)

vchuravy:
It looks like travis was cancelled on my latest PRs can you restart it?

vchuravy:
@pluskid Would be good to merge from my end.

Issue #126, opened 2016-08-30
Fix deprecation warnings


ViralBShah:
Perhaps this is already fixed for 0.5 and only a tag is needed.

```
WARNING: `@windows_only` is deprecated, use `@static if is_windows()` instead
 in depwarn(::String, ::Symbol) at ./deprecated.jl:64
 in depwarn(::String, ::Symbol) at /Users/viral/Desktop/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in @windows_only(::Any) at ./deprecated.jl:489
 in include_from_node1(::String) at ./loading.jl:426
 in include_from_node1(::String) at /Users/viral/Desktop/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in evalfile(::String, ::Array{String,1}) at ./loading.jl:442 (repeats 2 times)
 in cd(::##2#4, ::String) at ./file.jl:59
 in (::##1#3)(::IOStream) at ./none:13
 in open(::##1#3, ::String, ::String) at ./iostream.jl:113
 in eval(::Module, ::Any) at ./boot.jl:234
 in eval(::Module, ::Any) at /Users/viral/Desktop/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in process_options(::Base.JLOptions) at ./client.jl:239
 in _start() at ./client.jl:318
 in _start() at /Users/viral/Desktop/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
while loading /Users/viral/.julia/v0.5/MXNet/deps/build.jl, in expression starting on line 20
```

Cc @ranjanan

vchuravy:
Are you on current master of MXNet.jl? I am hesitant to tag a new release until I have addressed the remaining documentation issues.

ViralBShah:
I believe that is what I had with the latest METADATA. I don't think this is urgent in any way.

vchuravy:
@pluskid I think we can tag a new release and close this.

pluskid:
Tagged, waiting for merging.

ViralBShah:
On the latest tagged MXNet, I still see:

```
julia> using MXNet
INFO: Precompiling module MXNet.

WARNING: deprecated syntax "[a=>b for (a,b) in c]".
Use "Dict(a=>b for (a,b) in c)" instead.

WARNING: deprecated syntax "[a=>b for (a,b) in c]".
[...]

PR #127, opened 2016-08-31
cleanup docs a bit more

vchuravy:
Another round of cleanups.

### DIFF
--- a/docs/src/api/optimizer.md
+++ b/docs/src/api/optimizer.md
@@ -1,12 +1,21 @@
 # Optimizers
 
 ```@autodocs
-Modules = [MXNet.mx]
+Modules = [MXNet.mx, MXNet.mx.LearningRate, MXNet.mx.Momentum]
 Pages = ["optimizer.jl"]
 ```
 
 ## Built-in optimizers
 
-```@contents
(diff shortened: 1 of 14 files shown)


vchuravy:
This should fix most of the docs except for https://github.com/JuliaDocs/Documenter.jl/issues/247, https://github.com/JuliaDocs/Documenter.jl/issues/228 and https://github.com/JuliaDocs/Documenter.jl/issues/227

@pluskid can we tag a new release for #126 or do you want to wait on #123?

pluskid:
Do you know if Julia v0.5 is going to be released soon? If it is very close, we could try to wait for that and try to fix most of 0.5 related warnings before tagging a new version. If is it not going to happen soon, then I'm OK with tagging after #126.

vchuravy:
The remaining v0.5 warnings are coming from https://github.com/JuliaLang/Formatting.jl/issues/23 and the fact that `Compat` can't handle `Dict([x=>x for x ... ])`. The dict issue we can't directly fix until we drop support for v0.4, but it is fairly subdued.

If I would venture a guess v0.5 should be released within the next two weeks, as long as the current RC has no show-stoppers.

pluskid:
Thanks! Sounds good. I think we can make a quick new version.

Issue #132, opened 2016-09-06
mx.sum no longer works with SymbolicNode


MikeInnes:
Since 3fac768636662e7befc95764fe32996c8e477494 `mx.sum == Base.sum`; looks like the change made MXNet override Base's version rather than creating its own, so the method is still there but never called.

MikeInnes:
Looks like this is fixed with 9e3a60a56120fa75ad0291ea0cbf81af146c7fc4

PR #133, opened 2016-09-06
Fix mx.sum regression

MikeInnes:
Possible fix for #132. I think this should recover the original behaviour, although it may be better to simply add `::SymbolicNode` dispatch to the currently untyped symbolic methods.


vchuravy:
Good catch. I will add a PR to fix the dispatch.

@pluskid Should we stop extending base methods and require the use of `mx.sum`? Especially with the planned proliferation of methods in MXNet proper that follow `numpy` semantics instead of Julia semantics.

pluskid:
@MikeInnes Thanks!
@vchuravy Yes, I think it is good idea to require using `mx.` prefix before everything gets super confusing. We should emphasize this in the doc.

vchuravy:
@pluskid The only problem with that is that we then lose the niceness of
```
A = mx.Variable(:A)
B = mx.Variable(:B)
C = A + B'
```

vchuravy:
I think this was mostly done with #134?

PR #134, opened 2016-09-06
Positional arguments in symbolic API need type restriction.

vchuravy:
This fixes the issue noted in #132, but is orthogonal to #133


pluskid:
@vchuravy Thanks! If I understand correctly, we should still merge #133 , right?

vchuravy:
Yes, but we will need to decide what happens with the current list of manual imports. https://github.com/dmlc/MXNet.jl/blob/master/src/MXNet.jl#L17-L18
The reason why I put that auto-import in was to prevent issues like  #42

pluskid:
@vchuravy My understanding is as long as we use proper type annotation for the mx functions, auto-import should be fine.

vchuravy:
That is true, we probably should then also to auto-import for NDArray and remove the manual imports? 

We could even try to generate a list of the auto-imported functions and put that as a note in the manual.

pluskid:
@vchuravy Sounds like a good idea!

PR #135, opened 2016-09-06
remove mistaken + in REQUIRE

vchuravy:
Thanks @tkelman for noticing this

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.4+
+julia 0.4
 Compat 0.8.4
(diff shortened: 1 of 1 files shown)


vchuravy:
@pluskid v0.0.9 should probably also include #134

pluskid:
@vchuravy I think it is going to be very messy to force changing the existing tag. I will make a tag of v0.0.10 or v0.1.0 after some clean up.

vchuravy:
Sounds good to me.

PR #139, opened 2016-09-07
improve buildscript

vchuravy:
This cleans the buildscript a bit up.

@pluskid This should definitely run through the CI first. I tested it locally against prebuild versions of Julia, but I am not a 100% sure that `MakeTargets` works as I expect.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -44,19 +44,25 @@ if !libmxnet_detected
   end
 
   if blas_vendor == :unknown
-    error("Julia is build with an unkown blas library ($blas_path).\n Automatic building of libmxnet is not yet supported.")
-  elseif blas_vendor != :openblas64 && blas_vendor != :openblas
-    warn("Unsure if we can build against $blas_vendor.")
+    info("Julia is build with an unkown blas library ($blas_path).")
+    info("Attempting build without reusing the blas library")
+    USE_JULIA_BLAS = false
+  elseif !(blas_vendor in (:openblas, :openblas64))
+    info("Unsure if we can build against $blas_vendor.")
+    info("Attempting build anyway.")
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -44,48 +44,69 @@ if !libmxnet_detected
   end
 
   if blas_vendor == :unknown
-    error("Julia is build with an unkown blas library ($blas_path).\n Automatic building of libmxnet is not yet supported.")
-  elseif blas_vendor != :openblas64 && blas_vendor != :openblas
-    warn("Unsure if we can build against $blas_vendor.")
+    info("Julia is build with an unkown blas library ($blas_path).")
```
tkelman: built?

deps/build.jl
```diff
@@ -44,48 +44,69 @@ if !libmxnet_detected
...
-          `cp lib/libmxnet.so $_libdir`
+          if USE_JULIA_BLAS
+          MakeTargets("USE_BLAS=$blas_name -j$(nprocs())", env=Dict(
+                      "MSHADOW_LDFLAGS" => blas_path,
+                      "MSHADOW_CFLAGS" => ilp64,
+                     ))
+          else
+            `make -j$(nprocs())`
+          end
+          `rm $_libdir/libmxnet.so`
```
pluskid: `rm -f`, otherwise the command will fail if the file does not exist.

pluskid:
merged locally and tested after fixing the comments above.

pluskid:
@vchuravy It seems Travis CI still don't like it. It says symbol `cblas_sgemm` not found. Maybe the openblas library is not properly linked.

tkelman:
The generic linux binary renames all symbols to avoid colliding with non ILP64 libraries also being loaded

PR #142, opened 2016-09-08
Fixes deps/build.jl on Travis

vchuravy:
The problem is that we need to specify `MSHADOW_LDFLAGS` and
`MSHADOW_CFLAGS`. The way MXNet is currently setup we can't change these
from the terminal. This solves the problem by using `sed` to change
`mshadow/make/mshadow.mk`. But if mshadow would change git could throw
an error on updating. So we remove the changes before an update and add
them afterwards again. This has the drawback that users can't modifty
that file. (But those users should maintain their own build and use
`MXNET_HOME`.)

This also eagerly deletes `usr/lib/libmxnet.so` So that we actually
trigger a rebuild on an update of the pkg.

The travis build for this can be found at https://travis-ci.org/vchuravy/MXNet.jl/builds/158598024

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -56,6 +56,7 @@ if !libmxnet_detected
   end
 
   blas_name = blas_vendor == :openblas64 ? "openblas" : string(blas_vendor)
+  MSHADOW_LDFLAGS = "MSHADOW_LDFLAGS=-lm $blas_path"
 
   #--------------------------------------------------------------------------------
   # Build libmxnet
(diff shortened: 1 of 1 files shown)


vchuravy:
cc: @tqchen This fixes the Julia tests on MXNet proper. Sorry for breaking them so soon after re-enabling them.

tqchen:
@vchuravy The issue i pointed was another issue, on bash syntax error after reaching to the julia tests part, maybe you want to take a look of that

vchuravy:
@tqchen Just noticed that see https://github.com/dmlc/mxnet/pull/3260

Issue #11, opened 2015-10-29
[TODO] More optimizers


pluskid:
Currently we have SGD and Adam. Should be relatively simple to add more, like RMSProp, etc. Some references:

* http://cs231n.github.io/neural-networks-3/#update
* https://github.com/Lasagne/Lasagne/blob/master/lasagne/updates.py
* https://github.com/fchollet/keras/blob/master/keras/optimizers.py

tbreloff:
I would consider adding AdaMax as well... it's very briefly referenced in the Adam paper, but I think it's actually better.

pluskid:
@tbreloff Yes that would be great! I somehow get swamped in something else and was not able to add those shortly. But you are very welcome to contribute AdaMax (or other optimizers that you think is useful).

Mike43110:
Is there interest in a Genetic Algorithm backend? I can start a pure Julia one once I have played around with MXNet more. I don't have experience however with adding the CUDA backend.

vchuravy:
@Mike43110 Optimizers don't require CUDA backends. What kind of genetic algorithm would you like to add?

Mike43110:
I am planning on starting with just a basic one. Will have elitism as a flag and multiple crossover and mutation types. Will then move on to trying an adaptive GA and parallel implementations if performance is improved. My previous tests did not however show significant improvement with processor parallelism. Threading at a later stage along with testing of shared arrays still needs to be done.

I have not attempted using a PSO or DE based evolution of a network yet. But, I should be able to add those as well.

PR #159, opened 2016-12-27
New Optimizers

Arkoniak:
Implemented some new optimizers, as was requested in [https://github.com/dmlc/MXNet.jl/issues/11](https://github.com/dmlc/MXNet.jl/issues/11)

* AdaDelta
* AdaGrad
* AdaMax
* RMSProp
* Nadam

Also, made slight change in ADAM code, as written in [https://arxiv.org/pdf/1412.6980v8.pdf](https://arxiv.org/pdf/1412.6980v8.pdf): "Note that the efficiency of algorithm 1 can, at the expense of clarity, be improved upon by changing the order of computation, e.g.  by replacing the last three lines in the loop"

New optimizers can be tested in examples/mnist/mlp-test.jl

### DIFF
--- a/docs/src/api/optimizer.md
+++ b/docs/src/api/optimizer.md
@@ -19,3 +19,20 @@ Modules = [MXNet.mx]
 Pages = ["optimizers/adam.jl"]
 ```
 
+### AdaGrad
+```@autodocs
+Modules = [MXNet.mx]
+Pages = ["optimizers/adagrad.jl"]
+```
+
+### AdaDelta
+```@autodocs
+Modules = [MXNet.mx]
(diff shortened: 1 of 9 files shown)


pluskid:
Thanks a lot!

vchuravy:
This is great! Thank you.

vchuravy:
We should probably add correctness tests for these optimizers.

pluskid:
Wouldn't `mnist_fit_and_predict` function as testing? Or do you mean unit test? I'm not sure what would be good way to unit test optimizers, apart from implementing then again in another way?

PR #162, opened 2017-01-03
ENH: added verbosity training option to control printouts

sglyon:
This adds another option to `TrainingOptions` to control the verbosity of printed messages during training.

The default behavior hasn't changed.

I made fairly arbitrary decisions regarding what the cutoffs were for verbosity levels `n` and `n+1`, so if anyone has comments on how that can be improved I'm all ears.

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -191,7 +191,7 @@ end
     for copying mini-batches of data. Since there is no concern about convergence in prediction, it is better
     to set the mini-batch size as large as possible (limited by your device memory) if prediction speed is a
     concern.
-    
+
     For the same reason, currently prediction will only use the first device even if multiple devices are
     provided to construct the model.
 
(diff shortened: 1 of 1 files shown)


Arkoniak:
May be add all statistics output as a callback, like `do_checkpoint`? Then it could be more flexible, like for example, you'll be able to fine tune which eval metrics show/not show, how they could be shown (choose format string as an argument of this callback), where output to (stdout/file/both or some other option) and so on. I especially like the idea of outputting to file, because it'll be an easy task to setup small web server that could chart your progress online. And leave only general verbosity - 0/1 in your current setting.

spencerlyon2:
I totally agree with what you've said. It sounds much more flexible. 

That being said, I don't think I know the current code base well enough or have the time to implement what you are suggesting. 

What I was going for here was a quick way to avoid flooding my console with printouts that I didn't need.

pluskid:
Thanks! I guess more flexible solution might be to customize the logger interface and use different command `info`, `debug`, etc. to do logging or with different verbosity if that is supported by logging.jl. But this PR is good for now. Thanks!

sglyon:
Thanks for reviewing/merging

PR #165, opened 2017-01-07
Fix scalar operations in NDArray

Arkoniak:
Fixed conversion of julia params to libmxnet. Added more tests in test/ndarray.jl

### DIFF
--- a/src/ndarray.jl
+++ b/src/ndarray.jl
@@ -928,6 +928,13 @@ end
   ACCEPT_EMPTY_MUTATE_TARGET = (1 << 2)
 )
 
+function _julia_to_mx_param(val :: Any)
+  string(val)
+end
+function _julia_to_mx_param(val :: Real)
+  @sprintf("%e", val)
+end
+
 # Import corresponding math functions from base so the automatically defined libmxnet
 # functions can overload them
 import Base: sqrt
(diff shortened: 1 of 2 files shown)


test/unittest/ndarray.jl
```diff
@@ -7,8 +7,8 @@ using ..Main: rand_dims, reldiff
 ################################################################################
 # Test Implementations
 ################################################################################
-function rand_tensors{N}(dims::NTuple{N, Int})
-  tensor = rand(mx.MX_float, dims)
+function rand_tensors{N}(dims::NTuple{N, Int}, float_type=mx.MX_float)
```
vchuravy: A more Julian version of this would be:
```
rand_tensors{N}(dims::NTuple{N, Int}) = rand_tensors(mx.MX_float, dims)
function rand_tensors{N,T}(::Type{T}, dims::NTuple{N, Int})
```

Arkoniak:
Thank you, changed as you suggested. it is always nice to know new ways of doing things!

src/ndarray.jl
```diff
@@ -928,6 +928,16 @@ end
...
 
+function _julia_to_mx_param(val :: Any)
+  string(val)
+end
+function _julia_to_mx_param(val :: Float16)
+  string(val)
+end
+function _julia_to_mx_param(val :: Real)
+  @sprintf("%e", val)
+end
```
iblis17: Can I merge these function with `dump_mx_param`, which locate in `base.jl`?

PR #169, opened 2017-01-11
Fix of bugs in nnvm branch

Arkoniak:
This PR contains following changes and improvements. 
1. Added `@testset` wrappers in tests. Now tests are running till the end of all test cases, and show nice table of corresponding number of errors in each test set. It's more convenient than falling on first error.
2. Improved conversion from Float to String. It turns out, that simple "%e" rounds numbers, which in turn causes failures in different tests like ndarray clip. Here is gist, which shows that new format setting gives one to one correspondence between number and it's string representation: https://gist.github.com/Arkoniak/56abec14e110688adec5ed857d443685 Also lowered accuracy for Float16 tests.
3. Fixed bugs in julia 0.4 version of script. First of all, in this version julia was unable to convert Array{Union{}} to Ptr{Ptr{Void}} (it is needed when args is empty), secondly, julia 0.4 has weird bug, that it ignores wrapping of array in another array. Manual construction of Array{Array{MX_handle}} solved this problem.
4. Removed deprecation warnings in julia 0.5/0.6 which appeared in Dict([for]) construction. Changing of list comprehension to map function removed this warning and almost has no speed impact: https://gist.github.com/Arkoniak/ba09fbe189d6818d3edd3d89de63f5d7
5. Added function `/(arg :: Real, self :: SymbolicNode)` which is needed for julia 0.6 version.

I've tested this commit in travis and it passes 0.4 and 0.5 version, both on linux and mac: https://travis-ci.org/Arkoniak/MXNet.jl/builds/191010902
There are two errors in 0.6, but they are not nnvm related:

1. In `visualize` tagged version of JSON  fails during parse. This is known issue, it is fixed on master, but not tagged yet: https://github.com/JuliaIO/JSON.jl/issues/183
2. Also loader functions for MNIST fails with 'world age' error. It is known too: https://discourse.julialang.org/t/help-calling-a-function-defined-from-expressions/1340/4?u=skoffer and requires some redefinition of corresponding functions.

So, I think, after this PR nnvm branch is finally ready to be merged with master.


Arkoniak:
By the way, sometimes travis jobs just fails without any obvious reason. Sometimes MXNet.jl just throws some unexplainable error and dies, sometimes not. I wasn't able to reproduce such a behaviour locally, and I guess it is related to interoperation of julia, c++ and travis environment. May be someone with experience in this fields can help...

vchuravy:
> By the way, sometimes travis jobs just fails without any obvious reason. Sometimes MXNet.jl just throws some unexplainable error and dies, sometimes not. I wasn't able to reproduce such a behaviour locally, and I guess it is related to interoperation of julia, c++ and travis environment. May be someone with experience in this fields can help...

If you have a log like that best back it up as a gist so that one of use can take a look at it :)

src/model.jl
```diff
@@ -384,8 +384,8 @@ function fit(self :: FeedForward, optimizer :: AbstractOptimizer, data :: Abstra
 
   train_execs = Array(Executor, num_dev)
   for i = 1:num_dev
-    data_shapes = Dict([k => tuple(v[1:end-1]...,length(slices[i])) for (k,v) in provide_data(data)])
-    label_shapes = Dict([k => tuple(v[1:end-1]...,length(slices[i])) for (k,v) in provide_label(data)])
+    data_shapes = Dict(map((x) -> x[1] => tuple(x[2][1:end-1]...,length(slices[i])), provide_data(data)))
```
vchuravy: This should work until we can drop support for `v0.4` and we can just use generators

Arkoniak:
> If you have a log like that best back it up as a gist so that one of use can take a look at it :)
Plenty of those, unfortunately.
1. In this PR (linux julia 0.4): https://travis-ci.org/dmlc/MXNet.jl/jobs/191109397
2. Linux julia 0.5: https://travis-ci.org/Arkoniak/MXNet.jl/jobs/191099815
3. MacOS julia 0.4 (failed without any warnings at all): https://travis-ci.org/Arkoniak/MXNet.jl/jobs/190772048
4. MacOS julia 0.5(unable to compile MXNet.jl): https://travis-ci.org/Arkoniak/MXNet.jl/jobs/190772049

pluskid:
@Arkoniak Thanks for your efforts! I have now merged the nnvm branch into master. We will wait for some time to see if there is any bugs to be fixed, and then we shall tag a new release to catch up with the libmxnet development.

Issue #167, opened 2017-01-10
Custom output such as exponential function?


JockLawrie:
Hi there,

I am trying to define the activation of the last layer as the exponential function. If x is the input vector to a node in the last layer, the output of the node would be exp(w*x + b). Is this possible?

Thanks!
Jock

pluskid:
I think it should be possible with basic operations like matrix multiplication, plus, and exp.

JockLawrie:
I'm not sure how to get this working. As an example, suppose I have a net with 1 hidden layer of 10 nodes and an output layer of 4 nodes. Suppose that the activation of output layer should be `:softrelu`. The code below doesn't work, partly because I haven't specified where `mx.Variable(:label)` comes into the net. What am I missing? And how might I replace softrelu with the exponential function?

Thanks in advance.

```julia
net = @mx.chain mx.Variable(:data) =>

    mx.FullyConnected(name = :fc1_in, num_hidden = 10) =>
    mx.Activation(name = :fc1_out, act_type = :relu) =>

    mx.FullyConnected(name = :fc2_in, num_hidden = 4) =>
    mx.Activation(name = :fc2_out, act_type = :softrelu)
```

Arkoniak:
You can use symbolic calculations to get what you need.

```
using MXNet

# net without activation layer
net = @mx.chain mx.Variable(:data) =>

    mx.FullyConnected(name = :fc1_in, num_hidden = 10) =>
    mx.Activation(name = :fc1_out, act_type = :relu) =>

    mx.FullyConnected(name = :fc2_in, num_hidden = 4)

# Same net with exponential activation
net_out = mx.exp(net, name = :fc2_out)

# Here outputs of two nets are joined together, for easier comparison
net = mx.Group(net, net_out)

println(mx.list_arguments(net))       # arguments = input data + hidden layers weights
println(mx.list_outputs(net))             # outputs, since we use grouped net, we have two outputs: before and after activation

# some random data for forward propagation. Since we are not going to train model, no labels are needed
x = rand(Float32, 10, 2)
data = mx.ArrayDataProvider(:data => x, batch_size=2)

model = mx.FeedForward(net)

# usually you do not use this function directly, it is called internally from train function
mx.init_model(model, mx.UniformInitializer(), data=(10, 2))

# This is forward pass with some random weights. We get two arrays, before and after exponential activation 
res = mx.predict(model, data)

# And we can check, that everything is fine
[...]

Issue #166, opened 2017-01-10
Is creation of custom operators supported?


philtomson:
In the MXNet docs there's a section on how to create new custom operators (Layers) and the doc there suggests this can be done in the "front-end" language - Python is suggested, but I'm wondering if this is doable in Julia as well? The document talks about using CustomOp for doing this, but I'm not seeing that under mx.

Specifically, their example code for Python shows:

class Softmax(mx.operator.CustomOp)
...

I'm not seeing an mx.operator.CustomOp when I search around in the Julia REPL.

vchuravy:
This is not yet possible in Julia, but it is something that I am working on.

philtomson:
@vchuravy What's involved in doing this? I have some time, I might be able to help out.

Arkoniak:
I suppose there was a long discussion in this thread, that may be of interest to you: https://github.com/dmlc/MXNet.jl/pull/16
Also there was some discussion in https://github.com/dmlc/MXNet.jl/issues/39#issuecomment-180260704 with link to source and example, that can be fixed.

vchuravy:
Both #16 and #130 are the wrong way of doing this :) The problem is that async callbacks into Julia are tricky to get right (no GIL to help us out) and if you do get them right they are pretty slow, limited and brittle. I discussed this with pluskid over the summer and the right way of doing this is to bring the module system to MXNet.jl (https://github.com/dmlc/MXNet.jl/issues/129#issuecomment-244780450). 

I am now back on working on this so I hope to make progress in the next weeks.

Arkoniak:
@vchuravy Can you tell what are the main guidelines for implementing this api and what schedule do you have in mind? I've started to implement module api as well, it's still heavy WIP, but you can see it here: https://github.com/Arkoniak/MXNet.jl/tree/mx_module/src

It is based on https://github.com/dmlc/MXNet.jl/tree/vc/module, port of python code (incomplete) and reuse of FeedForward model. If you have any progress of your own or you know main points that should be taken care of, it would be nice to cooperate, because this module feature is a real must have. May be this topic deserve it's own issue.

vchuravy:
@Arkoniak My work is on https://github.com/oist/MXNet.jl/tree/vc/module and I am starting to work on this again today and aim to finish it soon (tm).

vchuravy:
@Arkoniak I rebased my work and if you are keen to also work on this I would appreciate PRs against vc/module here.

Issue #129, opened 2016-09-05
Bucketing in MXNet.jl


vchuravy:
I think we are missing the infrastructure necessary to implement bucketing? http://mxnet.readthedocs.io/en/latest/how_to/bucketing.html

@ranjanan This would be helpful for you.

pluskid:
@vchuravy Yes, we could add that in the current FeedForward interface. Or it would be better if we could add the whole Module API in the Python side to Julia. I have wanted to do this, but never found a chunk of continuous time slot to work on it.

vchuravy:
You mean http://mxnet.readthedocs.io/en/latest/packages/python/module.html ? I can make an attempt at wrapping the underlying interface.

pluskid:
@vchuravy Yes, exactly! You are very welcome to if you have time!

ZacCranko:
Any progress on wrapping the Module interface?

Issue #77, opened 2016-04-13
Any plans to add custom loss functions?


Simplex-AP:
The python version of MXNet has the capability to define custom loss functions, but this is missing from the Julia version. Are there any plans to add it?

vchuravy:
Depends on #39 and see my work in #16. The ability to define custom loss functions requires layers written in python and is what inspired me to work on it. I have plans to work on this but it will need to wait till julia v0.5

pluskid:
Another option which might be easier to implement (though involves quite a lot refactoring but not technically difficult) is the new module system in the mxnet python side. See dmlc/mxnet#1868 for example. With the new module system, a hybrid symbolic and imperative module can be used. The computation graph is built using symbolic nodes, while the loss function is written directly in Python.

Arkoniak:
There is a [discussion](https://github.com/dmlc/mxnet/issues/3368) In R branch of MXNet. They propose to use MakeLoss function. I've tried to implement their solution:
* [linear regression](https://gist.github.com/Arkoniak/5b9b33f399daae726bfe581cbd37fee1)
* [logistic regression](https://gist.github.com/Arkoniak/4ae1910dcb210e2b0db7fbd7027261ba)

Well, generally speaking it works, but it looks ugly. it would be nice, if anyone with deep knowledge of MXNet could help.
* Because NN + custom loss layer has loss output as a result, it is not clear, how to calculate other eval metrics during training.
* For the same reason, i had to keep two different neural networks, one for fit, another one for predict. May be there is better solution? Default output layers like SoftmaxOutput and others do not have this problem.
* Even if there is need to keep two different networks, it would be nice to abstract loss layer, for future reuse. I guess it should be bind/simple_bind somehow to base net, but wasn't able to how to use this syntax.

PR #173, opened 2017-01-16
[WIP/RFC] Initial support for modules

vchuravy:
This is one of the bigger features that the Python frontend has that we are currently missing.
My goals with this are to support several different use-cases:

- Bucketing: #129 
- Custom operators in Julia: #39, #77, #166, #167 
- For optimal performance I plan on integrating this with CUDAnative.jl for Julia v0.6 so that custom operators can be executed on the GPU

So far I have been working on a straight port of the Python functionality, but I would be interested in a bit more design discussion.

PRs against this branch are very welcome.


Arkoniak:
Thank you, it would be nice to have modules api implemented. I have couple of preliminary questions
1. Now `Nullable{...}` should be used instead of `Union{Void, ...}`?
2. May be make organization of modules in the same way as optimizers? I mean, move AbstractModule to "src/module.jl" and all implementations in "src/modules/...". it looks more natural to general structure of MXNet.jl.
3. Should `SymbolModule` support `AbstractDataProvider`? There is slight inconvenience, that `AbstractDataProvider` has its own functions for data/label names, so user will be forced to enter same information twice, moreover dataprovider's and module data_names can conflict. May be it would be better to infer names from `DataProvider` but unfortunately it is only possible during `bind`.
4. Are there any reasons to put modules in their own namespace? At first glance new functions shouldn't conflict with previous code, and `mx.Module.SymbolModule` looks redundant. But may be I am missing something important.

vchuravy:
1. Yes, `Nullable` is preferred. 
2. Will do that on the next push.
3. Will have to think about that.
4. Right now it is to make sure that things won't conflict and we can reexport the important bindings later once the interface has stabilised.

vchuravy:
Rebased to include #179.

vchuravy:
@pluskid If you have the time, it would be nice to get your opinion of the implementation so far.

pluskid:
@vchuravy Thanks a lot! Currently I did not have time to go through it very carefully. But quickly scanning it and it looks good. 

Regarding SequentialModule in another PR, maybe for `output_shapes`, instead of using Dict, we still keep vector of name, shape pairs?

vchuravy:
Some thoughts I had over the last days:

* `@pluskid` and I had a conversation in #153 on not doing automatic data parallelism in `SymbolModule` e.g. not using `ExecutorGroup`, what we could do instead is to simplify `SymbolModule` and provide a `DataParallelModule` wraps a module and parallelises it over a set of devices. This would split the complexity of `ExecutorGroup` and might make the API simpler.
* For sequential modules the wiring of input to output needs to be done during bind and there should be no need to pass data into backwards and forwards.

pluskid:
@vchuravy I agree that the current implementation is a bit complicated due to auto data par. `DataParallelModule` might be a good idea. I would support it if it solve the issue. Though maybe we need to inspect and design a bit more carefully. For example, data par implementation needs to make sure the gradients are aggregated via the same `KVStore`.

iblis17:
😱 why close this?

vchuravy:
I don't have any time to devote to this and I don't want the illusion of progress,

PR #178, opened 2017-01-19
Executor Group and bugfixes

Arkoniak:
This pr is still work in progress, and since my knowledge of julia is rather limited, will be glad to change code in julian way.

This pr has following main features:
1. Fixed small bugs in original PR, so it now compiles.
2. Made tests for symbol module in unittest directory, and changed accordingly original test-module.jl. So, tests now are running from runtest as usual, and one can run only symbol module tests with `julia test-module.jl`
3. Prototype implementation of executor group, which is a mixture of port of python code and some parts were borrowed from `FeedForward` model.
4. Removed partially `Nullable`. Main idea, that in most cases it is enough to have an empty vector, so `isnull` check is excessive. It is important, though, to have `fixed_param_names` as `Nullable`, because in current implementation if `fixed_param_names` is null, then model attrs are used to determine which gradient to use. If `fixed_param_names` is not null then its value is used to determine which gradient to use. Also `is_train` in forward is Nullable, because there is additional behaviour in case if `is_train` is null. On the other hand I can't figure out, when specific behviour is needed in case of `data_shapes` and the like.
5. Added special `DataBatchProvider`, which makes available syntax of the form
```julia
for batch in eachdatabatch(provider)
  # Here batch is of DataBatch type, so it is possible to do things like
  info(batch.data)

  # Or, since, DataBatch contains all needed data it is possible to use shorter form of forward
  # in python-like style
  forward(symbol_module, batch)
end
```
but I am unhappy with implementation of `forward(x::AbstractModule, y::DataBatch) = forward(x, DataBatchProvider(), y)`, which is base on the fact, that `get_data(provider::AbstractProvider, ::DataBatch)` do not requires `provider`, so it can be anything. It works, but I think better solution exists.

6. Removed partially type definitions in `executor.jl:copy_params_from`. Since `copy` function operates as `copy(::Union{NDArray, Array}, ::Union{NDArray, Array})` this removal makes it able to pass julia Arrays during parameter initialization. Alternative way, as I can see is defining something like
`copy_params_from(..., ::Union{Void,Dict{Base.Symbol,Union{NDArray, Array{Real}}}} ,....)` but this looks really ugly. May be it makes sense to use typealias here, but since this type checking is not used in type dispatch, I think it is ok to remove it completely.
7. Also added `return self` in all init functions, so it will be possible to do chaining.

### DIFF
--- a/src/MXNet.jl
+++ b/src/MXNet.jl
@@ -37,6 +37,7 @@ include("kvstore.jl")
 
 include("callback.jl")
 include("model.jl")
+include("executor-group.jl")
 include("module/Module.jl")
 
 include("visualize.jl")
--- a/src/executor-group.jl
+++ b/src/executor-group.jl
@@ -1,0 +1,225 @@
+"""
+    AbstractExecutorGroup
+Executor group is a convenient tool for managing a group of executors.
+"""
+abstract AbstractExecutorGroup
+
+function forward(self::AbstractExecutorGroup, data_provider :: AbstractDataProvider,
+                 data_batch :: AbstractDataBatch, is_train)
+  throw(MethodError(forward, (self, )))
+end
+
+type DataParallelExecutorGroup <: AbstractExecutorGroup
+  symbol :: SymbolicNode
+  context :: Vector{Context}
+  execs :: Vector{Executor}
+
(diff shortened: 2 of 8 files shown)


vchuravy:
Cool, Thank you.

PR #179, opened 2017-01-22
Medium and High Level api for SymbolModule

Arkoniak:
Implemented medium and high level api, examples (linear regression) in unittest/symbol-module.jl

Not implemented:
1. Callbacks in `fit`, `score`
2. `KVStore` support, right now kvstore defaults to `nothing`.
3. `reset` in `fit`, `score`, because it involves rewriting of `AbstractDataProvider`

Problems:
1. Tests coverage is very low
2. Docs is a mess. Some style guides needed
3. Code should be cleaned: fields in types should contain definitions, function definitions can be improved.
4. Due to 1. it probably contains some unexpected bugs.

I am switching now to the implementation of `SequentialModule`, because it will be useful in checking `grad_outputs`.

### DIFF
--- a/src/executor-group.jl
+++ b/src/executor-group.jl
@@ -16,7 +16,10 @@ type DataParallelExecutorGroup <: AbstractExecutorGroup
 
   data_shapes :: Dict{Symbol, Tuple{Vararg{Int}}}
   label_shapes :: Dict{Symbol, Tuple{Vararg{Int}}}
+
   for_training :: Bool
+  slices :: Vector{UnitRange{Int}}
+  batch_size :: Int
 
   shared_group :: Nullable{DataParallelExecutorGroup}
   inputs_need_grad :: Bool
@@ -127,7 +130,7 @@ function DataParallelExecutorGroup(symbol::SymbolicNode, context::Vector{Context
 
   return DataParallelExecutorGroup(
     symbol, context, execs,
-    data_shapes, label_shapes, for_training,
+    data_shapes, label_shapes, for_training, slices, batch_size,
     shared_group, inputs_need_grad, fixed_param_names, grad_req, freeze_idx,
     data_arrays, label_arrays, param_arrays, grad_arrays, aux_arrays,
     input_grad_arrays, arg_params, aux_params)
@@ -157,6 +160,21 @@ function forward(self:: DataParallelExecutorGroup, data_provider :: AbstractData
    # TODO add callbacks here
 end
 
+# TODO Add description
+backward(self::DataParallelExecutorGroup, out_grads::Void) = backward(self, NDArray[])
+backward(self::DataParallelExecutorGroup, out_grads::NDArray) = backward(self, [out_grads])
+function backward(self::DataParallelExecutorGroup, out_grads::Vector{NDArray}=Vector{NDArray}())
+  @assert self.for_training, "re-bind with for_training=True to run backward"
+  
+  for (i, exec) in enumerate(self.execs)
+    out_grad_slices = NDArray[]
+    for grad in out_grads
+      push!(out_grad_slices, copy(grad, self.context[i]))
+    end
+    backward(exec, out_grad_slices)
+  end
+end
+
 """
 		set_params!(self::DataParallelExecutorGroup, arg_params, aux_params)
 
@@ -173,6 +191,47 @@ function set_params!(self::DataParallelExecutorGroup,
     copy_params_from(exec, arg_params, aux_params, allow_extra_params=allow_extra_params)
   end
 end
+
+##
+# Utility
+##
+
+update_params(self::DataParallelExecutorGroup, updater, update_on_kvstore, kvstore::Void = nothing) = update_params(self, updater, update_on_kvstore, Nullable{KVStore}())
+update_params(self::DataParallelExecutorGroup, updater, update_on_kvstore, kvstore::KVStore) = update_params(self, updater, update_on_kvstore, Nullable(kvstore))
+function update_params(self::DataParallelExecutorGroup, updater, update_on_kvstore, kvstore::Nullable{KVStore})
+  num_dev = length(self.context)
+  for idx = 1:length(self.param_names)
+    #= if isa(self.grad_arrays[i][1], Void) =#
+    #=   continue =#
+    #= end =#
+    if in(idx, self.freeze_idx)
+      continue # Skip parameter update entirely
+    end
+    if !isnull(kvstore)
+      kvstore = get(kvstore)
+      # push gradient, priority is negative index
+      push!(kvstore, idx, self.param_arrays[idx], priority=-idx)
+      if update_on_kvstore
+        # pull back the weights
+        pull!(kvstore, idx, self.param_arrays[idx], priority=-idx)
+      else
+        # pull back the sum-ed gradients, to the same locations
+        pull!(kvstore, idx, self.grad_arrays[idx], priority=-idx)
+      end
+    end
+
+    if !update_on_kvstore
+      # manual updating
+      for i_dev = 1:num_dev
+        # create a fake index, so that the updater create states
+        # for different param AND different devices, TODO(mli)
+        # use a better solution later
+        fake_idx = idx * num_dev + i_dev
+        get(updater)(fake_idx, self.grad_arrays[idx][i_dev], self.param_arrays[idx][i_dev])
+      end
+    end
+  end
+end 
 
 ##
 # Internals
--- a/src/executor-group.jl
+++ b/src/executor-group.jl
@@ -14,8 +14,8 @@ type DataParallelExecutorGroup <: AbstractExecutorGroup
   context :: Vector{Context}
   execs :: Vector{Executor}
 
-  data_shapes :: Dict{Symbol, Tuple{Vararg{Int}}}
-  label_shapes :: Dict{Symbol, Tuple{Vararg{Int}}}
+  data_shapes :: Vector{Tuple{Vararg{Int}}}
+  label_shapes :: Vector{Tuple{Vararg{Int}}}
 
   for_training :: Bool
   slices :: Vector{UnitRange{Int}}
@@ -36,6 +36,8 @@ type DataParallelExecutorGroup <: AbstractExecutorGroup
 
   arg_params :: Dict{Symbol, NDArray}
   aux_params :: Dict{Symbol, NDArray}
+  param_names :: Vector{Symbol}
+  aux_names :: Vector{Symbol}
 end
 function DataParallelExecutorGroup(symbol::SymbolicNode, context::Vector{Context},
            data_shapes, data_names, label_shapes, label_names, for_training, inputs_need_grad,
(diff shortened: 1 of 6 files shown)


src/executor-group.jl
```diff
@@ -100,14 +105,15 @@ function DataParallelExecutorGroup(symbol::SymbolicNode, context::Vector{Context
...
+    arg_shapes_dev, out_shapes_dev, aux_shapes_dev = infer_shape(symbol; data_shapes_dev..., label_shapes_dev...)
+    @assert(!isa(arg_shapes_dev, Void), "Information not enough to perform complete shape inference")
+    arg_arrays = NDArray[zeros(shape, context[i]) for shape in arg_shapes_dev]
     grad_arrays = Dict{Symbol,NDArray}()
-    aux_arrays = NDArray[zeros(T, shape, context[i]) for (shape, T) in zip(aux_shapes, aux_types)]
+    aux_arrays = NDArray[zeros(shape, context[i]) for shape in aux_shapes_dev]
 
-    # Process arguments
-    arg_info = zip(arg_names, arg_shapes, arg_types)
+    shapes = zip(arg_names, arg_shapes_dev)
```
vchuravy: Here and above seem to be rebase/merge errors

src/executor-group.jl
```diff
@@ -22,9 +22,12 @@ type DataParallelExecutorGroup <: AbstractExecutorGroup
   context :: Vector{Context}
   execs :: Vector{Executor}
 
-  data_shapes :: Dict{Symbol, Tuple{Vararg{Int}}}
-  label_shapes :: Dict{Symbol, Tuple{Vararg{Int}}}
+  data_shapes :: Vector{Tuple{Vararg{Int}}}
+  label_shapes :: Vector{Tuple{Vararg{Int}}}
```
vchuravy: Any particular reason to go back to vectors from dicts?
We should just speed up the dict pathway to remove the splatting penalty.

src/executor-group.jl
```diff
@@ -178,15 +182,31 @@ function forward(self:: DataParallelExecutorGroup, data_provider :: AbstractData
...
 end
 
+# TODO Add description
+backward(self::DataParallelExecutorGroup, out_grads::Void) = backward(self, NDArray[])
+backward(self::DataParallelExecutorGroup, out_grads::NDArray) = backward(self, [out_grads])
+function backward(self::DataParallelExecutorGroup, out_grads::Vector{NDArray})
+  @assert(self.for_training, "re-bind with for_training=true to run backward")
+  
+  for (i, exec) in enumerate(self.execs)
+    out_grad_slices = NDArray[]
```
vchuravy: Can we allocate these in `bind`? and use type information.

src/executor-group.jl
```diff
@@ -196,15 +216,161 @@ function set_params!(self::DataParallelExecutorGroup,
...
+
+  # XXX: there is a possibiilty, that label arrays lie in different
+  # context than cpu_output_arrays. It should be checked and labels 
+  # should be copied to corresponding context
+  cpu_output_arrays = get_outputs(self)
+  update!(eval_metric, get_label(provider, batch), cpu_output_arrays)
+end
+
+"""
+		get_outputs
```
vchuravy: Whitespace

src/executor-group.jl
```diff
@@ -196,15 +216,161 @@ function set_params!(self::DataParallelExecutorGroup,
...
+    for i in 1:length(block)
+      @inplace w .+= copy(block[i], cpu())
+    end
+    @inplace w ./= length(block)
+    copy!(aux_params[name], w)
+  end
+end
+
+"""
+
```
vchuravy: Missing signature

src/executor-group.jl
```diff
@@ -178,15 +182,31 @@ function forward(self:: DataParallelExecutorGroup, data_provider :: AbstractData
    # TODO add callbacks here
 end
 
+# TODO Add description
+backward(self::DataParallelExecutorGroup, out_grads::Void) = backward(self, NDArray[])
```
vchuravy: Do we need `out_grads`?

src/executor-group.jl
```diff
@@ -196,15 +216,161 @@ function set_params!(self::DataParallelExecutorGroup,
...
-# Internals
+# Utility
 ##
 
+update_params(self::DataParallelExecutorGroup, updater, update_on_kvstore, kvstore::Void = nothing) = update_params(self, updater, update_on_kvstore, Nullable{KVStore}())
+update_params(self::DataParallelExecutorGroup, updater, update_on_kvstore, kvstore::KVStore) = update_params(self, updater, update_on_kvstore, Nullable(kvstore))
+function update_params(self::DataParallelExecutorGroup, updater, update_on_kvstore, kvstore::Nullable{KVStore})
+  num_dev = length(self.context)
+  for idx = 1:length(self.param_names)
+    #= if isa(self.grad_arrays[i][1], Void) =#
```
vchuravy: type-instabilities like that should be avoided and `Nullable` used instead

src/executor-group.jl
```diff
@@ -196,15 +216,161 @@ function set_params!(self::DataParallelExecutorGroup,
...
+* `arg_params` : Dict{Symbol, Vector{NDArray}}. Target parameter arrays
+* `aux_params` : Dict{Symbol, Vector{NDArray}}. Target aux arrays
+
+# Notes
+This function will inplace update the NDArrays in arg_params and aux_params.
+"""
+function get_params!(self::DataParallelExecutorGroup, arg_params::Dict{Symbol, NDArray}, 
+                    aux_params::Dict{Symbol, NDArray})
+  for (name, block) in zip(self.param_names, self.param_arrays)
+    w = empty(size(block[1]))
```
vchuravy: This needs to take type-inference into account. In effect there should be no un-typed `empty`, `zero`, or `one`

src/executor-group.jl
```diff
@@ -244,3 +410,5 @@ function get_grads(symbol, param_names, arg_names, data_names, inputs_need_grad,
 
   return grad_req_dict, freeze_idx
 end
+
+_merge_multi_context(outputs) = [concatenate(tensors, always_copy=false) for tensors in outputs]
```
vchuravy: Since this is so simply, maybe inline the function in the two places where it is used? Makes it easier to understand what is going on in these places.

src/module/Module.jl
```diff
@@ -1,6 +1,7 @@
 module Module
 import ....MXNet: mx
 import ..mx: DataBatch, AbstractDataProvider, AbstractDataBatch, DataBatchProvider
+import ..mx: SymbolicNode, NDArray, Context, Executor, list_arguments, infer_shape, GRAD_NOP, AbstractExecutorGroup, list_outputs, DataParallelExecutorGroup, KVStore, OptimizationState, ADAM, UniformInitializer, set_params!, AbstractOptimizer, get_updater, update_params, provide_data, provide_label, AbstractEvalMetric, StubProvider, init, copy!, concatenate, eachdatabatch, reset!, Accuracy
```
vchuravy: this needs a line-break,

src/module/Module.jl
```diff
@@ -119,18 +120,27 @@ end
...
+A Dict of (name, shape) pairs specifying the data inputs to this module. 
 """
 function data_shapes(self :: AbstractModule)
   throw(MethodError(data_shapes, (self,)))
 end
 
 """
+    label_shapes(AbstractModule)
+
+A Dict of (name, shape) pairs specifying the label inputs to this module.  If this module does not accept labels -- either it is a module without loss function, or it is not binded for training, then this should return an empty Dict.
```
vchuravy: Line-break.

src/io.jl
```diff
@@ -17,6 +17,9 @@ Normally this involves defining:
 """
 abstract AbstractDataProvider
 
+type StubProvider <: AbstractDataProvider
```
vchuravy: Why is `StubProvider` necessary?

src/module/symbol_module.jl
```diff
@@ -64,12 +62,12 @@ type SymbolModule <: AbstractModule
 end
 
 function SymbolModule(symbol::SymbolicNode;
-                data_names = [:data], label_names = [:softmax_label],
+                      data_names = [:data],
+                      label_names = [:softmax_label],
                 context = [mx.cpu()], fixed_param_names = nothing)
```
vchuravy: indentation is off

src/module/symbol_module.jl
```diff
@@ -224,68 +258,117 @@ function init_optimizer(self::SymbolModule; optimizer::AbstractOptimizer=ADAM(),
...
   #= end =#
 
+  self.optimizer_initialized = true
+
   return self
 end
 
-# TODO add description
+"""
+		get_outputs
```
vchuravy: indentation is off

src/module/symbol_module.jl
```diff
@@ -224,68 +258,117 @@ function init_optimizer(self::SymbolModule; optimizer::AbstractOptimizer=ADAM(),
...
+	should merge the collected results so that they look like from a single
+	executor.
+
+# Returns
+If `merge_multi_context` is `true`, it is like `[out1, out2]`. Otherwise, it
+is like `[[out1_dev1, out1_dev2], [out2_dev1, out2_dev2]]`. All the output
+elements are `NDArray`.
+"""
+function get_outputs(self::SymbolModule, merge_multi_context::Bool=true)
+	@assert isbinded(self) && isinitialized(self)
```
vchuravy: indentation is off.

src/module/symbol_module.jl
```diff
@@ -224,68 +258,117 @@ function init_optimizer(self::SymbolModule; optimizer::AbstractOptimizer=ADAM(),
...
-                   self.exec_group.param_arrays,
-                   self.exec_group.grad_arrays,
-                   updater=self.updater,
-                   num_device=length(self.context))
-  end
+  update_params(self.exec_group, self.updater, self.update_on_kvstore, self.kvstore)
+end
+
+"""
+		update_metric()
```
vchuravy: indentation

vchuravy:
I would like to merge this soon :) Do you think you have time to address the review comments? Otherwise I would do them myself.

vchuravy:
I am in the process of merging this locally.

src/executor-group.jl
Arkoniak: Yes, they are used in `SequentialModule` to pass information between submodules.

Arkoniak: Oh, I guess I forget to add final `return out_grad_slices` in this function.

src/executor-group.jl
Arkoniak: Yes you are right, it has no sense.

src/executor-group.jl
Arkoniak: Sorry, my knowledge of Julia is not that good, to understand what is needed. But if you show correct piece of code, than i can follow it in the future. The same applies to previous comment.

src/executor-group.jl
Arkoniak: Agreed. It started as rather long python function, and simplified to one line. It should be refactored away now.

src/io.jl
Arkoniak: It is used in all functions, of the kind `function f(..., y::DataBatch, ...) = f(..., StubProvider(), y, )`. I thought, that construction `DataBatchProvider() = new(); DataBatchProvider(provider) = new(provider)` looks uglier, comparing to additional type. If better way to resolve this problem exists, or if `DataBatchProvider` syntax is ok, that's fine by me.

src/executor-group.jl
vchuravy: Ah that makes sense. Then we should update the module API definition to make a requirement for all `<: AbstractModule`.

src/executor-group.jl
vchuravy: No worries I will take care of that :)

vchuravy:
My feeling is that we should polish the implementation for `SymbolModule` so that it is well tested and documented. Then we can merge my current PR and add additional PRs for `Pipeline` and `NativeModule`.

Arkoniak:
Oh. It looks like `PipelineModule` is actually the same as `SequentialModule`. `SequentialModule` is almost ready and I was thinking to make PR. I can rename it to `PipelineModule` and make corresponding changes if needed.

vchuravy:
You can also keep SequentialModule as a name :)

PR #181, opened 2017-01-24
Sequential module

Arkoniak:
This is somewhat pre alpha implementation of SequentialModule (a.k.a Pipeline Module). Havn't tested it fully yet, suppose it's time to cover with tests and docs. So this PR is mainly for review.

What has been done:
1. Some tests restructure - added new test file, `test-module.jl` now has the same structure as `runtests.jl`, except it covers only `*module.jl` files.
2. Had to restructure partially `bind` and `init_params`, because it quickly became hard to maintain these long arguments definition. I am afraid it's a mess of dispatch calls now :-(
3. Tested up to `bind`.

Current problems: 
1. I am not very happy with current syntax. For example, existence of default `label_names` in `SymbolModule` means, that one has to remove them explicitly in every new layer declaration.
2. Also usage of `Dict` in `ExecutorGroup` for `data_shapes` and `label_shapes` turns out to be a problem, because functions like `bind` can receive either `Vector` or `Dict` as an input, and some additional logic required to process it.
3. Also, if `output_shapes` of a `Module` is a `Vector` than it can be easily plugged into next `Module`, because `Vector` preserve ordering. On the other hand `Dict` do not preserve order, so additional work is required, either one have to carefully construct Modules and check that `data_names` of next module are the same as `output_names` of the previous module, or some sort of joining should be done, like supply name correspondence in different modules.

### DIFF
--- a/src/executor-group.jl
+++ b/src/executor-group.jl
@@ -22,8 +22,8 @@ type DataParallelExecutorGroup <: AbstractExecutorGroup
   context :: Vector{Context}
   execs :: Vector{Executor}
 
-  data_shapes :: Vector{Tuple{Vararg{Int}}}
-  label_shapes :: Vector{Tuple{Vararg{Int}}}
+  data_shapes :: Dict{Symbol, Tuple{Vararg{Int}}}
+  label_shapes :: Dict{Symbol, Tuple{Vararg{Int}}}
 
   for_training :: Bool
   slices :: Vector{UnitRange{Int}}
@@ -160,6 +160,9 @@ function DataParallelExecutorGroup(symbol::SymbolicNode, context::Vector{Context
   else
     input_grad_arrays = []
   end
+
+  data_shapes = Dict(name => shape for (name, shape) in zip(data_names, data_shapes))
+  label_shapes = Dict(name => shape for (name, shape) in zip(label_names, label_shapes))
 
   return DataParallelExecutorGroup(
     symbol, context, execs,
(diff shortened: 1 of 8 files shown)


pluskid:
Thanks! Adding support for this should enable writing simple loss function or even layers in Julia easily without having to wait for Julia to properly support multi-thread in callbacks (though the performance might be slower).

1. I think the default label names are to support several image classification examples. I have no objection if there are better ways to handle this.
2. 
3. agreed.

Arkoniak:
Well, after some considerations, I have following ideas.
1. Move label's definition (names and shapes) from Module declaration to `bind` function. I've checked and there is no need to define labels before `bind`. This would make api more simple: in `SequentialModule` it would be possible to set labels correctly, and wouldn't make any difference in all other modules.
2. Actually it is possible to restore order from corresponding `data_names` and `output_names`, so in the end Dicts are fine. So, binding procedure is the following
2.1. If length(output) == 1 then auto-wiring is forced, because we know exactly how inputs and outputs correspond.
2.2 if length(output) > 1 then we test if Set(output_names) == Set(data_names). If yes, then we just pass `output_shapes` in the order of `data_names`
2.3. If previous two fails, but user set option `auto_wire = true` in metas, then we pass `output_shapes` in the order of `output_names`
2.4. As last resort, if all above fails, user may set option `join = Dict{Symbol, Symbol}` which set how `output_names` corresponds to `data_names`, and shapes will be passed in that order.

It's longer to write than to code, I'll send commit to this PR soon.

pluskid:
LGTM. Shall we merge this?

PR #188, opened 2017-01-27
Using WebRequest function to download the JSON file

bmharsha:
`wget` was failing to download mxnet.json , so replaced them with `Invoke-WebRequest`. After this change, MXNet.jl works on Windows, `Pkg.test("MXNet")` passes too.

Tested on
1) Windows Server 2012
2) Windows server 2008 SP1, R2


vchuravy:
Thank you for testing! I tried this on Windows 10 and it worked there.

PR #194, opened 2017-01-29
Improve GPU detection

vchuravy:
This is a quality of life improvement to our libmxnet build process. I have tested this on my local Linux box, but I would appreciate wider testing (especially on Windows)

Fixes: #175, #143

PR #196, opened 2017-01-29
remove usr/setupenv.cmd because it is too invasive

vchuravy:
addresses @tkelman comment in https://github.com/dmlc/MXNet.jl/pull/194#issuecomment-275901350

If this is ok I will merge asap and put out a new patch release.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -40,7 +40,6 @@ if !libmxnet_detected
 
     run(download_cmd(base_url, "mxnet_base.7z"))
     run(`7z x mxnet_base.7z -y -ousr`)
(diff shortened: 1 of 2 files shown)


tkelman:
Windows wasn't supported at all on any releases prior to 0.2.0, right? So hopefully it's fair to be patient and take a bit more time to work out a Julia-local setup method.

vchuravy:
No, we didn't support automatic installation on Windows prior to v0.2.0.

pluskid:
Thanks @vchuravy! And no, Windows automatic building was not supported before.

PR #206, opened 2017-02-23
get_name function 

MikeInnes:
Sometimes it's useful to get the name of a symbolic node; this just makes that easy.


src/symbolic-node.jl
```diff
@@ -200,6 +200,27 @@ function set_attr(self :: SymbolicNode, key :: Symbol, value :: AbstractString)
...
+
+    julia> y = mx.FullyConnected(x, num_hidden = 128)
+    julia> mx.get_name(y)
+    :fullyconnected0
+"""
+function get_name(self :: mx.SymbolicNode)
+    name = Ref{mx.char_p}(0)
+    success = Ref(0)
+    @mxcall(:MXSymbolGetName, (MX_handle, Ref{char_p}, Ref{Int}), self.handle.value, name, success)
+    @assert success[] != -1
```
pluskid: It seems symbol-node created by `Group` does not have names. Could you please change this assert into some more informative thing? Like returning an empty symbol or null, or raise an exception with more helpful information?

test/unittest/symbolic-node.jl
```diff
@@ -100,6 +101,7 @@ function test_attrs()
   @test isnull(mx.get_attr(conv, :b))
   @test get(mx.get_attr(conv, :a)) == "a"
   @test get(mx.get_attr(conv, :π)) == "π"
+  @test isa(mx.get_name(conv), Symbol)
```
pluskid: Could you add a test for a grouped symbol to reflect the behavior you specified in this situation? Thanks!

src/symbolic-node.jl
MikeInnes: I'm not familiar with `Group`, can you give me an example of how one would be created? Is there any way to test if a node is a `Group` so that we can simply tell the user they don't have names in an error message?

pluskid: Hi @MikeInnes ! You can create a group in the following way:
```
julia> a = mx.Variable(:a)
MXNet.mx.SymbolicNode(MXNet.mx.MX_SymbolHandle(Ptr{Void} @0x00007fa45c4c0160))

julia> b = mx.Variable(:b)
MXNet.mx.SymbolicNode(MXNet.mx.MX_SymbolHandle(Ptr{Void} @0x00007fa45e882290))

julia> c = mx.Group(a, b)
MXNet.mx.SymbolicNode(MXNet.mx.MX_SymbolHandle(Ptr{Void} @0x00007fa45ec179d0))
```

I believe that you will simply get `success` being -1 when invoking the API if the underlying symbolic-node is actually a group. But maybe you can try it and confirm.

MikeInnes: Thanks. As it turns out, it just segfaults. So it looks like we're stuck as-is, unless you have away to pre-check whether the node is a `Group`.

pluskid:
I see...... let's merge it then.

MikeInnes:
Thanks!

vchuravy:
@MikeInnes can you report that seqfault to upstream MXNet?

PR #211, opened 2017-03-09
add a new metric: NMSE

iblis17:


src/metric.jl
```diff
@@ -179,6 +179,47 @@ function reset!(metric :: MSE)
   metric.n_sample = 0
 end
 
+doc"""
+    NMSE
+
+Normalized Mean Squared Error
+
+```math
+\sum_i (\frac{label_i - pred_i}{label_i})^2
```
vchuravy: Can you also add a reference? I am not familiar with NMSE but looking around online I found https://rem.jrc.ec.europa.eu/RemWeb/atmes2/20b.htm which seems to be different from the definition used here.

iblis17: I think the way to do *normalization* depends on your context.
There are tons of ways to do it. I always do it case-by-case.
That's why I explicitly document the formula.

Let me explain my case: I'm training a network for option pricing. The option pricing problem is a regression problem (pirce predicting). I have lots of option contract on same target stock but different strike price. For example, there is a stock `S`; it's market price is 1000. And, there are two call option contracts with different strike price:

```
+--------+----------------+----------------+--------------+
|        | Strike Price   | Market Price   | Pred Price   |
+--------+----------------+----------------+--------------+
| Op 1   | 1500           |  100           | 80           |
+--------+----------------+----------------+--------------+
| Op 2   |  500           |  10            | 8            |
+--------+----------------+----------------+--------------+
```

So, obviously, I will calculate my normalized MSE with
```
((100 - 80)/100)^2   vs   ((10 -8)/10)^2
```
I will consider both of the pred prices got the same degree of error.

And.... The formula you provided is not suitable for me. I guess it's suitable for some kind of clustering problem, like mixtuer model.
If you want me to implement that formula in this PR, I can make it.

pluskid:
LGTM. @iblis17 Could you add your explanation of the use case to the doc? It would be great to have a reference to something if there is one.

src/metric.jl
vchuravy: No worries and excellent explanation. If other cases popup we can introduce different cases of this.

iblis17:
Sure, i added it.

PR #212, opened 2017-03-09
Fixed data attribute

steventhornton:
Removed the key for the data attribute to match examples.


vchuravy:
Travis error is timeout on the mac buildbots.

PR #217, opened 2017-03-18
switch to Plots.jl from pyplot

mcreel:
Plots.jl allows for different backends, including PyPlot. The PR makes the example less dependent on the user's set of installed packages.


examples/regression-example.jl
```diff
@@ -56,7 +56,4 @@ mx.fit(model, optimizer, eval_metric=mx.MSE(), trainprovider, eval_data=evalprov
 # obtain predictions
 plotprovider = mx.ArrayDataProvider(:data => ValidationInput, :label => ValidationOutput)
 fit = mx.predict(model, plotprovider)
-plot(ValidationOutput,fit',".")
-xlabel("true")
-ylabel("predicted")
-title("outputs: true versus predicted. 45º line is what we hope for")
+scatter(ValidationOutput,fit',w = 3, xlabel="true", ylabel="predicted", title="45º line is what we hope for", show=true)
```
iblis17: How about wrapping this long line?

PR #220, opened 2017-03-28
Update ACE metric

vchuravy:
* Takes eps parameter to prevent log(0) to blow up,
* Recognises learning of probability distributions.


pluskid:
It seems travis CI failed on ACE?

PR #224, opened 2017-03-30
Improvements to metrices

vchuravy:
Depends on dmlc/mxnet#5629. Series of fixes improving metrices.
* Introduction of `SeqMetric` and `NullMetric`. `SeqMetric` splits the the outputs of the network and `NullMetric`
* Improve MSE to handle multidimensional data
* Introduce threaded reduction for MSE and ACE drastically speeding up calculations

Important to note is that `nd_as_jl` introduced type-instabilities and moving it across a function barriers allows Julia to infer the function properly.

### DIFF
--- a/src/MXNet.jl
+++ b/src/MXNet.jl
@@ -16,6 +16,7 @@ if VERSION >= v"0.6.0-dev.1024"
 end
 
 using Formatting
+using Base.Threads
 
 # Functions from base that we can safely extend and that are defined by libmxnet.
 import Base: round, ceil, floor, cos, sin, abs, sign, exp, sqrt, exp, log, norm,
--- a/src/metric.jl
+++ b/src/metric.jl
@@ -351,11 +351,11 @@ function _update_single_output{T}(metric :: ACE, label :: Array{T}, pred :: Arra
   @assert size(label, ndims(label)) == size(pred, ndims(pred))
 
   # ACE kernel for 4d
-  work4d = function (low, high, label, pred)
+  work4d = function (low, high, label, pred, eps)
     result = 0.0
     for idx in low:high
       i, j, sample = ind2sub(size(label, 1, 2, 4), idx)
-      @inbounds target = Int(labels[i, j, 1, sample]) + 1 # klasses are 0...k-1 => julia indexing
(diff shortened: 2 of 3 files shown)


pluskid:
@vchuravy Great work! Thanks! Could you try to port one of the simpler metric implemented in the Python side by directly using the NDArray api to Julia and compare the performance with the Julia threaded implementation? https://github.com/dmlc/mxnet/blob/master/python/mxnet/metric.py

If the performance is similar, we might prefer one with a simpler syntax for users to customize.

vchuravy:
On my localwork station `NDMSE` & `MSE` have both a nearly immeasurable impact on performance and most of the time is spend on `load_labels` and `load_data`.  Which I guess are main the synchroinization points?

I tried to enable `async` copy from the Julia side (controllable via `UV_THREADPOOL_SIZE`), but that aswell has nearly no impact on performance (at least on my workstation I am trying now on an NVLINK system).

vchuravy:
@pluskid There are ~two~ three changes to this that I am thinking about, before polishing and merging this.

1. Introduce a trait allowing metrics to select between `NDArray` and `Array` inputs
2. Removing the multithreading: It is neat and leads performance improvements in direct comparison, but in the larger picture it is not as significant anymore.
3. Is there a way that we could prefetch/preload labels and data?

pluskid:
@vchuravy I recommend measure the overall time of training, say, for one full epoch. So the measurement is not confused by the async flavor of the engine.

What do you mean by pre-fetch of data and labels? If you mean data IO, then there are pre-fetchers already implemented at least for the RecordIO that are used in, for example, imagenet IO.

pluskid:
+1 for removing multi-threading unless it has a significant improvements on performance. Otherwise, it might make things too complicated to debug and to extend.

+1 for traits to dinstinguish between NDArray and Julia Array.

vchuravy:
Ok I moved all the independent parts out and the last remaining one is multithreading. Once #233 is merged I will do a performance comparison across prefetch vs multithreading, vs non of these two options

vchuravy:
Testing shows that there is a linear speedup for multithreading, but given that #233 improves the situation so much I don't think there is much benefit in making metrics more complicated.

Issue #205, opened 2017-02-23
ArrayDataProvider returns TypeError


machiningcentre:
ArrayDataProvider returns TypeError when using Images.jl.
```
julia> using Images
julia> using MXNet
julia> X = rand(10,10);
julia> p = mx.ArrayDataProvider(X)
ERROR: TypeError: #ArrayDataProvider#6688: in new, expected Array{Array{Float32,N},1}, got Array{Array{Float32,N},1}
 in #ArrayDataProvider#6688(::Int64, ::Bool, ::Int64, ::Int64, ::Type{T}, ::Array{Float64,2}, ::Array{Any,1}) at ~/.julia/v0.5/MXNet/src/io.jl:351
 in (::Core.#kw#Type)(::Array{Any,1}, ::Type{MXNet.mx.ArrayDataProvider}, ::Array{Float64,2}, ::Array{Any,1}) at ./<missing>:0
 in MXNet.mx.ArrayDataProvider(::Array{Float64,2}) at ~/.julia/v0.5/MXNet/src/io.jl:277
```
There is no problem without Images.jl.
```
julia> using MXNet
julia> X = rand(10,10);
julia> p = mx.ArrayDataProvider(X)
MXNet.mx.ArrayDataProvider(Array{Float32,N}[
Float32[0.415866 0.112956 … 0837056 0.398794; 0.590402 0.334468 … 0515418 0.200105; … ; 0.3765410.337045 … 0.677437 0.31439; 0.761557 0.601104 … 0.7393450.0729888]],Symbol[:data],Array{Float32,N}[],Symbol[],10,10,false,0.0f0,0.0f0,MXNet.mx.NDArray[mx.NDArray{Float32}(10,10)],MXNet.mx.NDArray[])
```

Julia version and the package status are as follows:
```
julia> versioninfo()                                        
Julia Version 0.5.0                                         
Commit 3c9d753 (2016-09-19 18:14 UTC)                       
Platform Info:                                              
  System: Linux (x86_64-pc-linux-gnu)                       
  CPU: Intel(R) Xeon(R) CPU E5-1650 v3 @ 3.50GHz            
  WORD_SIZE: 64                                             
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_                                    
  LIBM: libopenlibm                                         
  LLVM: libLLVM-3.7.1 (ORCJIT, haswell)                     
                                                            
[...]

PR #228, opened 2017-04-12
fix conflict with Images

vchuravy:
Works around #205 most likely explanation is that subtyping went off the rails...


vchuravy:
The mac travis failure is in building mxnet proper:

```
In file included from /Users/travis/.julia/v0.5/MXNet/deps/src/mxnet/dmlc-core/include/dmlc/memory.h:12:
/Users/travis/.julia/v0.5/MXNet/deps/src/mxnet/dmlc-core/include/dmlc/./thread_local.h:40:12: error: thread-local storage is not supported for the current target
    static thread_local T inst;
```

PR #223, opened 2017-03-29
Added initializers for Spatial Transformer Networks (STN) layers

facundoq:
Spatial Transformer Networks were added to MXNet back in July 2016 (https://github.com/dmlc/mxnet/pull/2547/commits/12cc483c1c235b14ef12ab7cd12574d4b12d1f6c). To use them, specialized initializers are needed for the weights in the location network and the bilinear filter.  I've followed the implementation for the python frontend. Not sure if these needs testing from the Julia side since the core c++ functionality is already being tested by through python api in the MXNet repo.

PR #229, opened 2017-04-13
fixes bilinear initializer following approach in #34

vchuravy:
x-ref: https://github.com/dmlc/MXNet.jl/pull/223#issuecomment-293745376

@facundoq Can you double check that this is correct?

 I wonder why we never merged the initialisier part of #34 (quite odd).

### DIFF
--- a/src/initializer.jl
+++ b/src/initializer.jl
@@ -48,29 +48,28 @@ function _init_loc_bias(self :: AbstractInitializer, name :: Base.Symbol, array 
 end
 
 function _init_bilinear(self :: AbstractInitializer, name :: Base.Symbol, array :: NDArray)
-  # ported from python version:
-  #weight = np.zeros(np.prod(arr.shape), dtype='float32')
-  #shape = arr.shape
-  #f = np.ceil(shape[3] / 2.)
-  #c = (2 * f - 1 - f % 2) / (2. * f)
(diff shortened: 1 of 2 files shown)


facundoq:
Looks fine! 

My fix (https://github.com/facundoq/MXNet.jl/blob/c73bedc5e049675d66b1f7776b91c1d825c0cfe2/src/initializer.jl#L62-L72) did the same thing using linear indices and `%` to get  which seemed kinda odd to me but I sticked with it just to follow the python version. 

```
  weight=mx.zeros(length(array))
  h,w,channels,n=size(array)

  f = ceil(w / 2.)
  c = (2. * f - 1. - f % 2.) / (2. * f)
  for i=1:length(weight)
    x = (i-1) %  w
    y = div((i-1) ,w) % h
    weight[i:i] = (1.0 - abs(x / f - c)) * (1.0 - abs(y / f - c))
  end
  array[:]=mx.Reshape(weight,shape=(n,channels,w,h))
```

Both generate the same filters:
``array[:,:,1,1]=
 0.0625  0.1875  0.1875  0.0625
 0.1875  0.5625  0.5625  0.1875
 0.1875  0.5625  0.5625  0.1875
 0.0625  0.1875  0.1875  0.0625
```

vchuravy:
Good! Please submit your fix for the LTN initialisier

PR #232, opened 2017-04-14
Async prefetching and loading 

vchuravy:
This uses Julia tasks to preload (currently) 4 batches onto the device.

`@threadcall` is a normal `ccall` that does not block the main Julia thread. 

```
help?> @threadcall
  @threadcall((cfunc, clib), rettype, (argtypes...), argvals...)

  The @threadcall macro is called in the same way as ccall but does the work in a different thread. This is useful
  when you want to call a blocking C function without causing the main julia thread to become blocked. Concurrency is
  limited by size of the libuv thread pool, which defaults to 4 threads but can be increased by setting the
  UV_THREADPOOL_SIZE environment variable and restarting the julia process.

  Note that the called function should never call back into Julia.
```

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -61,6 +61,20 @@ macro mxcall(fv, argtypes, args...)
   end
 end
 
+"Utility macro to @threadcall MXNet API functions"
+macro mxthreadcall(fv, argtypes, args...)
+  f = eval(fv)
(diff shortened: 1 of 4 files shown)


vchuravy:
I won't have time to investigate the error on travis. So let's shelve this for now.

PR #233, opened 2017-04-14
Improve type-stability for metrices

vchuravy:
`_update_single_output` needs to be type-stable, therefore move `@nd_as_jl` over the function barrier

### DIFF
--- a/src/metric.jl
+++ b/src/metric.jl
@@ -236,27 +236,30 @@ Requires that label and prediction have the same shape.
 """
 
 type MSE <: AbstractEvalMetric
-  mse_sum  :: Float64
+  mse_sum  :: Vector{NDArray}
   n_sample :: Int
 
-  MSE() = new(0.0, 0)
+  MSE() = new(Vector{NDArray}(), 0)
 end
 
-hasNDArraySupport(::MSE) = Val{false}()
+hasNDArraySupport(::MSE) = Val{true}()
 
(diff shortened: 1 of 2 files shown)


vchuravy:
### Timings on a M6000:
|  | NullMetric | Seq(MSE, MultiACE) |
| - | ---- | ---- |
| Before |  65.5915 | 85.7034 |
| After | 65.2526 | 66.4901 |

vchuravy:
@pluskid We can also drop the last two commits

src/metric.jl
```diff
@@ -200,36 +229,37 @@ end
...
+  metric.n_sample += length(label)
+  mse_sum = mx.sum(mx._PowerScalar(label - pred,scalar=2))
+  push!(metric.mse_sum, mse_sum)
+  return nothing
 end
 
 function get(metric :: MSE)
-  return [(:MSE, metric.mse_sum / metric.n_sample)]
+  # Delay copy until last possible moment
+  mse_sum = mapreduce(nda->copy(nda)[1], +, 0.0, metric.mse_sum)
```
pluskid: My understanding is that each `NDArray` in `mse_sum` at this point are actually scalars? *Maybe* using `asscalar` will be more efficient than the generic `copy` function. Otherwise, LGTM.

vchuravy: `asscalar` only exist on the python side and is uses the same mechanism. We could probably write a more effective mechanism with an `unsafe_load`.

pluskid: I see. OK. Sounds good then.

PR #238, opened 2017-04-24
Fixed SpatialTransformerNetwork (stn) loc_bias initializer. Added lenet-stn example

facundoq:
As discussed in  #229. The code in #223 had a small bug, I initialized the loc_bias using `array = [..]` instead of `array[:]=[..]`. Sorry this took so long, but training with an STN sometimes failed and I wanted to make sure I was doing it right. 

I've also added a usage example.


pluskid:
Thank you very much!

PR #239, opened 2017-04-29
explicitly specify 7z

musm:


musm:
https://github.com/JuliaLang/julia/pull/21540

deps/build.jl
```diff
@@ -43,7 +43,8 @@ if !libmxnet_detected
     run(`cmd /c copy "usr\\3rdparty\\openblas\\bin\\*.dll" "usr\\lib"`)
 
     run(download_cmd(package_url, "mxnet.7z"))
-    run(`7z x mxnet.7z -y -ousr`)
+    exe7z = joinpath(JULIA_HOME, "7z.exe")
```
pluskid: Why `joinpath` with `JULIA_HOME`? Is 7z.exe shipped with Julia binary on windows?

pluskid:
@musm Thanks a lot! LGTM. Can you clarify the minor question above?

musm:
7z is shipped with Julia in Windows and the default windows juliarc file adds JULIAHOME so that 7z is automatically added to path. The default juliarc behavior will go away in future version of Julia, see linked issue. So we will now need to explicitly specify the 7z path on *windows*

pluskid:
Thanks for the clarification!

PR #247, opened 2017-05-23
Fix #242

### LINKED ISSUE #242: UndefVarError when providing KVStore object to mx.fit
I'm trying to train one model by invoking `mx.fit` multiple times. To reduce overhead, I'm creating my KV store beforehand and passing it to the `fit` method.
```julia
    kvstore = mx.KVStore(:device)
    mx.fit(n.model, optimizer, train_provider,
           n_epoch=1,
           eval_metric=metric,
           callbacks=[training_curve],
           verbosity=0,
           kvstore=kvstore)
```

The fit method then throws an error saying:
```
UndefVarError: update_on_kvstore not defined
```

This seems to be a bug in `mx.fit` where the local variable `update_on_kvstore` is only created when a `Symbol` is passed, not a `KVStore`.

From model.jl
```julia
  # setup kvstore
  kvstore = opts.kvstore
  if isa(kvstore, Base.Symbol)
    opts.verbosity >= 2 && info("Creating KVStore...")
    kvstore, update_on_kvstore = _create_kvstore(kvstore, length(self.ctx), self.arg_params)
  end
```

holl-:
Previously, when providing a KVStore object to mx.fit, the method would
crash with an UndefVarError.
This is now fixed by moving the definition of update_on_kvstore from
_create_kvstore to fit.


pluskid:
@holl- Thanks a lot!

PR #250, opened 2017-05-30
Use `Sys.CPU_CORES` not `nprocs()` when doing a parallel `make`

staticfloat:
This was always passing `-j1`, and after building this package a few times today, I'd really like it to use all my cores.  :)


pluskid:
Thanks!

PR #252, opened 2017-05-31
Try to work around appveyor rate limits

vchuravy:
@pluskid is there an option in appveyor to decrypt secret environment variables also for PRs?


pluskid:
@vchuravy Sorry, what do you mean?

vchuravy:
Sorry, I was not thinking clearly. We can't do this since Appveyor won't allow use to decrypt encrypted variables on PR (and I was wondering if there is that is configurable) and this will break the build for local checkouts.

PR #255, opened 2017-06-09
Add extra check on CUDA library detection

joshbode:
Some machines return a non-zero pointer for `Libdl.dlopen("")` causing
MXNet dep build to incorrectly report presence of CUDA libraries.

This is possibly a bug in Julia - I'm not sure where the non-NULL pointer goes :)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -36,7 +36,7 @@ end
 
 HAS_CUDA = false
 let cudalib = Libdl.find_library(["libcuda", "nvcuda.dll"], CUDAPATHS)
-  HAS_CUDA = Libdl.dlopen_e(cudalib) != C_NULL
+  HAS_CUDA = !isempty(cudalib) && Libdl.dlopen_e(cudalib) != C_NULL
 end
 
 if !HAS_CUDA && is_windows()


vchuravy:
That sound definitely like a bug upstream. Can you post your `versioninfo`?

joshbode:
Running in a Linux VM (on macOS) using the official 0.5.2 Linux binaries:
```
julia> Libdl.dlopen("")
Ptr{Void} @0x00007f183b11a1a8

julia> versioninfo()
Julia Version 0.5.2
Commit f4c6c9d (2017-05-06 16:34 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM)2 Duo CPU     P7550  @ 2.26GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Penryn)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, penryn)
```

Same behaviour also observed natively under Linux on a colleague's machine.

Same issue _not_ present on macOS natively, interestingly.

joshbode:
Reported here: JuliaLang/julia/issues/22318

joshbode:
I don't think this is a bug upstream. From `man 3 dlopen` on Linux:
> If filename is NULL, then the returned handle is for the main program.

joshbode:
Thanks :)

Issue #258, opened 2017-06-19
Git error adding package


JosephBoyd:
I get the following error on Linux Mint 18 Cinammon

INFO: Installing Formatting v0.2.1
INFO: Installing MXNet v0.2.3
INFO: Building MXNet
INFO: Found a CUDA installation.
INFO: Attempting to Create directory /home/joe/.julia/v0.5/MXNet/deps/src
INFO: Directory /home/joe/.julia/v0.5/MXNet/deps/src already created
INFO: Attempting to Create directory /home/joe/.julia/v0.5/MXNet/deps/usr/lib
INFO: Directory /home/joe/.julia/v0.5/MXNet/deps/usr/lib already created
INFO: Attempting to Create directory /home/joe/.julia/v0.5/MXNet/deps/src/mxnet
INFO: Directory /home/joe/.julia/v0.5/MXNet/deps/src/mxnet already created
INFO: Changing Directory to /home/joe/.julia/v0.5/MXNet/deps/src/mxnet
INFO: Changing Directory to /home/joe/.julia/v0.5/MXNet/deps/src/mxnet
INFO: Changing Directory to /home/joe/.julia/v0.5/MXNet/deps/src/mxnet
M	cub
HEAD is now at 8713d25... [R][DOC] update R installation guide (#6457)
INFO: Changing Directory to /home/joe/.julia/v0.5/MXNet/deps/src/mxnet
fatal: reference is not a tree: 89de7ab20167909bc2c4f8acd397671c47cf3c0d
Unable to checkout '89de7ab20167909bc2c4f8acd397671c47cf3c0d' in submodule path 'cub'
=========================================================================================[ ERROR: MXNet ]==========================================================================================

LoadError: failed process: Process(`git submodule update --init`, ProcessExited(1)) [1]
while loading /home/joe/.julia/v0.5/MXNet/deps/build.jl, in expression starting on line 58

===================================================================================================================================================================================================

=========================================================================================[ BUILD ERRORS ]==========================================================================================

WARNING: MXNet had build errors.

 - packages with build errors remain installed in /home/joe/.julia/v0.5
 - build the package(s) and all dependencies with `Pkg.build("MXNet")`
 - build a single package by running its `deps/build.jl` script

===================================================================================================================================================================================================
INFO: Package database updated

joshbode:
[...]

PR #259, opened 2017-06-20
Added cblas_.tr.m functions to cblas include file

joshbode:
Upstream `mxnet` has started using the `[ds]tr[sm]m` functions (dmlc/mxnet@e852036), causing a build error for MXNet.jl, e.g.

```
src/operator/tensor/./la_op_inline.h:27:86: error: 'cblas_dtrmm' was not declared in this scope
src/operator/tensor/./la_op_inline.h:27:86: error: 'cblas_dtrsm' was not declared in this scope
src/operator/tensor/./la_op_inline.h:27:86: error: 'cblas_strmm' was not declared in this scope
src/operator/tensor/./la_op_inline.h:27:86: error: 'cblas_strsm' was not declared in this scope
```

This change adds entries for the missing functions to the custom `cblas.h` file used in the dep build.


pluskid:
Thanks!

joshbode:
Thank you for that :)

Given that the build/install for MXNet.jl is currently broken for the latest release (see #258) would it make sense to create a new release soon?
It would probably require upstream to create a new tag for the build to target, unless `libmxnet_curr_ver` could point to a specific commit from mxnet.

pluskid:
Yes, I agree it probably makes sense to make a new release soon, also because that Julia 0.6 is now released. 

But it might be difficult to quickly get a new tag in the upstream, as the upstream is now in the Apache incubator and things like releasing needs to be planned and discussed. It might be easier just to repoint `libmxnet_curr_ver` if possible.

PR #272, opened 2017-09-10
reshape on NDArray not working?

iblis17:
```julia
julia> x = mx.zeros(3, 4)

julia> y = mx.reshape(x, shape=(4, 3)) 
mx.NDArray{Float32}(3, 4)         

julia> size(y)
(3, 4)

julia> copy(y) 
3×4 Array{Float32,2}:   
 0.0  0.0  0.0  0.0 
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
```

### DIFF
--- a/src/ndarray.jl
+++ b/src/ndarray.jl
@@ -921,6 +921,31 @@ function save(filename::String, data::Dict{Base.Symbol,NDArray})
           filename, length(names), arrays, names)
 end
 
+import Base: reshape
+
+"""
+    reshape(arr::NDArray, dim...; reverse=false)
+    reshape(arr::NDArray, dim; reverse=false)
+"""
+reshape{N}(arr::NDArray, dim::NTuple{N, Integer}; reverse::Bool=false) =
+    _reshape(arr, dim, reverse)
+reshape{N}(arr::NDArray, dim::Vararg{Integer, N}; reverse::Bool=false) =
+    _reshape(arr, dim, reverse)
+
+@inline function _reshape{N}(arr::NDArray, dim::NTuple{N, Integer}, reverse::Bool)
+  op_handle = _get_cached_libmx_op_handle("reshape")
+  n_output = Ref(Cint(0))
+  hdls_ref = Ref{Ptr{MX_handle}}(C_NULL)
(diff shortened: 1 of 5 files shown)


vchuravy:
This might be also necessary for the Symbolic API?

iblis17:
ha, I do not check Sumbolic API yet. 
I'm reading [c_api.h](https://github.com/apache/incubator-mxnet/blob/470b56437290b33fef51b067c96fdce08cefa584/include/mxnet/c_api.h) line-by-line :p 
And still no idea how Symbolic related stuff works.

src/ndarray.jl
```diff
@@ -921,6 +921,25 @@ function save(filename::String, data::Dict{Base.Symbol,NDArray})
...
+import Base: reshape
+
+reshape{N}(arr::NDArray, dim::NTuple{N, Integer}) = _reshape(arr, dim)
+reshape{N}(arr::NDArray, dim::Vararg{Integer, N}) = _reshape(arr, dim)
+
+@inline function _reshape{N}(arr::NDArray, dim::NTuple{N, Integer})
+  h_ref = Ref{MX_handle}(C_NULL)
+  cint_dim = reverse(collect(Cint, dim))
+  @mxcall(:MXNDArrayReshape, (MX_handle, Cint, Ref{Cint}, Ref{MX_handle}),
+          arr, N, Ref(cint_dim), h_ref)
```
iblis17: I also found that there are a `reshape` operator support NDArray reshaping. what's the main difference? And... seems that (before this PR) we are using the `reshape` operator on both NDArray and SymbolicNode.

vchuravy: Looking at the two implementation the `reshape` operator seems to be more flexible and we should probably use them.

iblis17:
switched to operator already.
ready for review

pluskid:
I agree the difference between row-based indexing and column based indexing in Julia and C++ backend are causing some gray areas. Especially about the meaning of the arguments for operators like reshape, transpose / swapaxes, etc. This issue is made worse with the auto-generated doc from the C++ backend which usually focus on describing the Python behavior.

PR #275, opened 2017-09-10
build: remove FileRule on config.mk and fix CUDA config

iblis17:
* remove FileRule on config.mk, we always override it.
  Due to the following workflow is quite common:
    * user issue `Pkg.build("MXNet")`

    * then, build failed because the `CUDA_HOME` not set properly.

    * if the `FileRule` find `config.mk`,
      re-issuing `Pkg.build` won't change `config.mk`.
      User need to delete/modify `config.mk` manually

* fix CUDA config
    * the build_steps didn't being unfolded correctly.

    * consider following cases:

```julia
@build_steps begin
  if true
    `1`
    `2`
    `3`
    if true
      `42`
    end
  end
end
BinDeps.SynchronousStepCollection(Any[`42`], "", "")
```

vs

```julia
@build_steps begin
  if true
    @build_steps begin
      `1`
      `2`
      `3`
      if true
        `42`
      end
    end  # inner build_steps
  end  # if
end
BinDeps.SynchronousStepCollection(Any[`1`, `2`, `3`, `42`], "", "")
```

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -159,22 +159,30 @@ if !libmxnet_detected
           `git submodule update --init --recursive`
           `git -C mshadow checkout -- make/mshadow.mk`
           `make clean`
+          `cp ../../cblas.h include/cblas.h`
+
           `sed -i -s "s/MSHADOW_CFLAGS = \(.*\)/MSHADOW_CFLAGS = \1 $ilp64/" mshadow/make/mshadow.mk`
-        end
-        FileRule(joinpath(_mxdir, "config.mk"), @build_steps begin
-          ChangeDirectory(_mxdir)
+
+          # Copy config.mk, always override the file
           if is_apple()
             `cp make/osx.mk config.mk`
           else
             `cp make/config.mk config.mk`
           end
+
+          # Configure OpenCV
           `sed -i -s 's/USE_OPENCV = 1/USE_OPENCV = 0/' config.mk`
+
+          # Configure CUDA
           if HAS_CUDA
-            `sed -i -s 's/USE_CUDA = 0/USE_CUDA = 1/' config.mk`
-            if haskey(ENV, "CUDA_HOME")
-              `sed -i -s 's/USE_CUDA_PATH = NULL/USE_CUDA_PATH = $(ENV["CUDA_HOME"])/' config.mk`
+            @build_steps begin
+              `sed -i -s 's/USE_CUDA = 0/USE_CUDA = 1/' config.mk`
+              if haskey(ENV, "CUDA_HOME")
+                `sed -i -s "s@USE_CUDA_PATH = NONE@USE_CUDA_PATH = $(ENV["CUDA_HOME"])@" config.mk`
+              end
             end
           end
+
           # Force enable LAPACK build
           # Julia's OpenBLAS has LAPACK functionality already
           if FORCE_LAPACK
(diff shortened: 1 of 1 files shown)


iblis17:
:/ my CUDA build on MXNet master got this

```julia
julia> Libdl.dlopen(Pkg.dir("MXNet", "deps", "usr", "lib", "libmxnet.so"))
ERROR: could not load library "/home/iblis/.julia/v0.6/MXNet/deps/usr/lib/libmxnet.so"                        /home/iblis/.julia/v0.6/MXNet/deps/usr/lib/libmxnet.so: undefined symbol: __cudaRegisterFatBinary
```

iblis17:
https://stackoverflow.com/a/23936313
Adding `-lcudart` solved my problem and testing passed (though test cases is CPU only)

iblis17:
Then, I got this:
```julia
julia> mx.zeros((2, 3), mx.gpu())   
julia: symbol lookup error: /home/iblis/.julia/v0.6/MXNet/deps/usr/lib/libmxnet.so: undefined symbol: cublasCreate_v2
```
and julia crashed

iblis17:
I added the linker flag: `-Wl,--no-undefined` to check the undefined symbols

Finally, figured out this
```diff
diff --git a/Makefile b/Makefile
index e26bd4df..431ce3a0 100644
--- a/Makefile
+++ b/Makefile
@@ -268,7 +268,7 @@ ALL_DEP = $(OBJ) $(EXTRA_OBJ) $(PLUGIN_OBJ) $(LIB_DEP)       
 ifeq ($(USE_CUDA), 1) 
        CFLAGS += -I$(ROOTDIR)/cub
        ALL_DEP += $(CUOBJ) $(EXTRA_CUOBJ) $(PLUGIN_CUOBJ) 
-       LDFLAGS += -lcuda -lcufft
+       LDFLAGS += -lcuda -lcufft -lcublas -lcusolver -lcurand -lcudart
        SCALA_PKG_PROFILE := $(SCALA_PKG_PROFILE)-gpu          
 else
        SCALA_PKG_PROFILE := $(SCALA_PKG_PROFILE)-cpu
```

vchuravy:
Hm weird what happens if you try to build upstream MXNet directly without going through Julia script?

iblis17:
Actually, I built from MXNet directly after I ran into https://github.com/apache/incubator-mxnet/pull/7838, because Julia script will do something like `git checkout master`

iblis17:
PR filed https://github.com/apache/incubator-mxnet/pull/7856

iblis17:
rebased for macOS build

deps/build.jl
```diff
@@ -159,22 +159,30 @@ if !libmxnet_detected
...
+
+          # Configure CUDA
           if HAS_CUDA
-            `sed -i -s 's/USE_CUDA = 0/USE_CUDA = 1/' config.mk`
-            if haskey(ENV, "CUDA_HOME")
-              `sed -i -s 's/USE_CUDA_PATH = NULL/USE_CUDA_PATH = $(ENV["CUDA_HOME"])/' config.mk`
+            @build_steps begin
+              `sed -i -s 's/USE_CUDA = 0/USE_CUDA = 1/' config.mk`
+              if haskey(ENV, "CUDA_HOME")
+                `sed -i -s "s@USE_CUDA_PATH = NONE@USE_CUDA_PATH = $(ENV["CUDA_HOME"])@" config.mk`
```
vchuravy: I have to admit I don't know `sed` well enough. Have you tested this locally?

iblis17: yes, I tested it.

Let me show you some `sed` usage: we can use `@` as delimiter to deal with something like path.
```sh
└─[iblis@kaladbolg]% echo 'hello sed' | sed -e "s@hello@/path/contains/slash@"                      
/path/contains/slash sed
```

vchuravy: Okay cool. LGTM

vchuravy:
Is this still WIP?

iblis17:
I'm waiting for https://github.com/apache/incubator-mxnet/pull/7856 and https://github.com/apache/incubator-mxnet/pull/7838.
I'm also fine with merging this PR first.

Issue #88, opened 2016-04-28
Representing the network as a graph.


vchuravy:
Is there any way besides parsing the json to work with the network as a graph?

```julia
using MXNet
using LightGraphs
using JSON

net = @mx.chain mx.Variable(:data) =>
mx.Convolution(num_filter = 64, kernel = (3,3), pad = (1, 1), stride=(1,1)) =>
mx.LeakyReLU(act_type = :prelu) => 
mx.Convolution(num_filter = 128, kernel = (3,3), pad = (1, 1), stride=(1,1)) =>
mx.LeakyReLU(act_type = :prelu) => 
mx.Convolution(num_filter = 2, kernel = (3,3), pad = (1, 1), stride=(1,1)) =>
mx.LeakyReLU(act_type = :prelu) =>
mx.SoftmaxOutput(name=:softmax, multi_output=true)

js_net = JSON.parse(mx.to_json(net))
nodes = js_net["nodes"]
graph = DiGraph(length(nodes))
for (idx, node) in enumerate(nodes)
    for edge in node["inputs"]
        x, _ = edge
        add_edge!(graph, x+1, idx)
    end
end
```

I am using this right now to determine which parts of the network should be frozen.

vchuravy:
As an example of how this can be useful, let's say you want to freeze the first few layers of a network

```julia
"""
Find nodes that are step away from the root
"""
function findNodes(graph, step)
    Set(findNodes(graph, 1, step))
end
    
function findNodes(graph, n, step)
    # anchor 
    if step < 0
        return Int[]
    end
    
    # Adding ourselves here in case that we don't visit any future edges
    nodes = Int[n]
    
    for node in in_neighbors(graph, n)
        push!(nodes, node)
    end
    
    for node in out_neighbors(graph, n)
        result = findNodes(graph, node, step-1)
        append!(nodes, result)
    end
    
    return nodes
end 
```

for i in findNodes(graph, 4)
    nodes[i]["attr"] = Dict("grad"=>"freeze")
end

net = mx.from_json(JSON.json(js_net), mx.SymbolicNode)
```

But it is kinda awkward to go through JSON instead of walking the network directly and using `mx.set_attr`.

PR #276, opened 2017-09-11
sym-node: support `print` for debugging

iblis17:
Ref: https://github.com/apache/incubator-mxnet/blob/470b56437290b33fef51b067c96fdce08cefa584/include/mxnet/c_api.h#L933

example: 
```julia
julia> print(mx.dot(mx.Variable(:x), mx.Variable(:y)))
Symbol Outputs:
        output[0]=dot4(0)
Variable:y
Variable:x
--------------------
Op:dot, Name=dot4
Inputs:
        arg[0]=y(0) version=0
        arg[1]=x(0) version=0
```

### DIFF
--- a/src/symbolic-node.jl
+++ b/src/symbolic-node.jl
@@ -220,6 +220,51 @@ function get_name(self :: mx.SymbolicNode)
     return Symbol(unsafe_string(name[]))
 end
 
+import Base: print
+
+function print(io :: IO, sym :: SymbolicNode)
+  out = Ref{mx.char_p}(C_NULL)
+  @mx.mxcall(:MXSymbolPrint, (mx.MX_SymbolHandle, Ref{mx.char_p}), sym.handle, out)
+  print(io, unsafe_string(out[]))
+end
+
+print(sym :: SymbolicNode) = print(STDOUT, sym)
+
+"""
+    print([io :: IO], sym :: SymbolicNode)
+
+Print the content of symbol, used for debug.
+
+```julia
+julia> layer = @mx.chain mx.Variable(:data)           =>
+         mx.FullyConnected(name=:fc1, num_hidden=128) =>
+         mx.Activation(name=:relu1, act_type=:relu)
+MXNet.mx.SymbolicNode(MXNet.mx.MX_SymbolHandle(Ptr{Void} @0x000055b29b9c3520))
+
+julia> print(layer)
+Symbol Outputs:
+        output[0]=relu1(0)
+Variable:data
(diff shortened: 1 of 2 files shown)


src/symbolic-node.jl
```diff
@@ -220,6 +220,51 @@ function get_name(self :: mx.SymbolicNode)
     return Symbol(unsafe_string(name[]))
 end
 
+import Base: print
+
+function print(io :: IO, sym :: SymbolicNode)
+  out = Ref{mx.char_p}(C_NULL)
+  @mx.mxcall(:MXSymbolPrint, (mx.MX_SymbolHandle, Ref{mx.char_p}), sym.handle, out)
```
vchuravy: Who is responsible for deallocating the string?

iblis17: https://github.com/apache/incubator-mxnet/blob/470b56437290b33fef51b067c96fdce08cefa584/include/mxnet/c_api.h#L933

```
MXNET_DLL int MXSymbolPrint(SymbolHandle symbol, const char **out_str);
```

(I'm not skilled at C) 
Do I need to free the content of `const char**`?

iblis17:
rebased for macOS build

iblis17:
found that there is a debug api for executor as well, maybe I should rename it as well?
https://github.com/dmlc/MXNet.jl/blob/2c5120d139fdd259f2d1b161ec94eea323d0fc20/src/executor.jl#L205-L221

src/symbolic-node.jl
vchuravy: I think technically we are taking  ownership of this ptr so yeah we need to free it after we are done with it, (probably elsewhere as well.)

vchuravy:
The problem with the executor debug string is that is very large and not necessarily a useful representation. in #88 I convert the network to a digraph to get a more visual representation

iblis17:
So... you want to remove debug_str api of executor?
How about adding #88 via specifying MIME of show?

vchuravy:
I don't want to remove it, but it is not a useful output for just `print`. But yeah if we can "render" it nicely depending on if we are outputting text or are running in a Jupyter session that would be awesome.

iblis17:
I see.
Let's back to symbolic node's debugging print. Is it okay to have a `print` API for it? I prefer to keep the consistency, let user using the same API.

vchuravy:
Sure!

vchuravy:
Can you rebase this for CI?

iblis17:
rebased

PR #274, opened 2017-09-18
ndarray: porting Python's autograd

iblis17:
* https://mxnet.incubator.apache.org/tutorials/gluon/autograd.html
* https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/autograd.py


iblis17:
TODO:
* [ ] port feature: user-defined differentiable function
    * In Python, it provides the [class Function](https://github.com/apache/incubator-mxnet/blob/065adb3702c110af7b537799be3ec9c16c27a72b/python/mxnet/autograd.py#L292). I also curious about how this could be implemented in Julia. If we defined a abstract type like `CustomFunc`, how can we make it callable?
* [ ] test cases:
    * [ ] attach_grad
    * [ ] grad
    * [ ] mark_variables
    * [ ] get_symbol
    * [ ] record
    * [ ] pause
    * [ ] train_mode
    * [ ] predict_mode
    * [ ] backward
* [ ] leverage `grad_req_map` from #283 
* [ ] update `NEWS.md`

vchuravy:
We don't have custom operators yet so I don't thing we need user-defined differential functions, (it is actually a very hard problem to solve in Julia...)

iblis17:
In Python, [this](https://github.com/apache/incubator-mxnet/blob/065adb3702c110af7b537799be3ec9c16c27a72b/python/mxnet/autograd.py#L410) shows that we only need one c api: `MXCustomFunctionRecord`.
So.. IMO we still can have user-defined diff funcs.
Also, IIUC, `make_symbol` translates the style of NDArray (dynamic) to Sybolic (static). Maybe AutoGrad can reveal a another way to create custom layer via combining `make_symbol` and user funcs?

About implementation of user funcs, maybe meta programming can help, just my imagination:

```
@mx.registerfunc MyFunc  # create callable function for it
forward(f::Myfunc, ...) = ...
backward(f::Myfunc, ...) = ...
```

but not sure where is the pitfall of implementing it.

iblis17:
found a problematic case:
```julia
julia> x = mx.NDArray([1 2; 3 4])
       mx.attach_grad(x)
       
       y = mx.record() do
         1x .* x
       end
       
       mx.backward(y)

julia> copy(mx.grad(x))
2×2 Array{Int64,2}:
 1   4
 9  16
```

It's gradient shoud be same as the one in commit message.

vchuravy:
I would leave custom functions completely out of the picture for now. The problems and pitfalls are plenty and I am currently not convinced there is a good way to implement them (outside of #173, which is currently stalled).

In order to make customs function work you need to guarantee that they are *never* called asynchronously.

iblis17:
> In order to make customs function work you need to guarantee that they are *never* called asynchronously.

which doc of mxnet c api can I consult? still can not understande what is the issue of aynchronous (io?).

vchuravy:
See the discussion and linked abandoned PRs in #166

pluskid:
Thanks! About custom OP, I think the brief summary is that libmxnet has a async engine when executing operators, so the Julia callback will be called in a different thread than the main thread. Due to the current incomplete support for multi-thread GC in Julia, a callback being called in a different thread from the C++ side is not guaranteed to function properly.

iblis17:
For record, here is an API changes:
 - https://github.com/apache/incubator-mxnet/pull/7698

iblis17:
https://github.com/apache/incubator-mxnet/blob/7484437d21c5e1456c24bff31e6642cf0a47db4d/src/imperative/imperative.cc#L182-L197

I ran into this error with recent libmxnet master,
seems that we should get rid of any use of `op(...; out = ...)`

iblis17:
I think this PR is ready for review.
I will send another PR for doc and docstring.

iblis17:
I try to implement user-defined function.
will send a PR after this PR merged.
https://github.com/dmlc/MXNet.jl/tree/ib/nd-ad-customfunc

iblis17:
@pluskid need more time to review?

iblis17:
travis failure is macOS build timeout.

PR #280, opened 2017-09-18
build with CuDNN support

iblis17:
https://github.com/apache/incubator-mxnet/blob/master/make/config.mk#L58

Blocker: ~#275~

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -36,6 +36,8 @@ if is_unix()
 end
 
 HAS_CUDA = false
+HAS_CUDNN = false
+cudnnver = -1
 let cudalib = Libdl.find_library(["libcuda", "nvcuda.dll"], CUDAPATHS)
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -48,8 +49,19 @@ if !HAS_CUDA && is_windows()
...
+  let cudnnlib = Libdl.find_library("libcudnn", CUDAPATHS)
+    HAS_CUDNN = !isempty(cudnnlib) && Libdl.dlopen_e(cudnnlib) != C_NULL
+  end
+end
+
 if HAS_CUDA
   info("Found a CUDA installation.")
+  if HAS_CUDNN
+      # TODO: do more version check?
+      cudnnver = dec(ccall((:cudnnGetVersion, "libcudnn"), Csize_t, ()))
```
iblis17: CI faliure looks weird... is `HAS_CUDA` on travis is `true`?

Issue #286, opened 2017-10-01
Examples won't run in Julia v0.6


SimonEnsemble:
Julia v0.6
Ubuntu 16.05
[Latest from `Pkg.add("MXNet")`. But `Pkg.checkout("MXNnet")` gives: `ERROR: MXNet is dirty, bailing`...]

In `MXNet/examples/mnist/`:

`julia lenet.jl` yields:

```
ERROR: LoadError: MethodError: MXNet.mx.#Activation(::Array{Any,1}, ::MXNet.mx.#Activation) is ambiguous. Candidates:
  (::MXNet.mx.#kw##Activation)(::Array{Any,1}, ::MXNet.mx.#Activation, args::MXNet.mx.SymbolicNode...) in MXNet.mx
  (::MXNet.mx.#kw##Activation)(::Array{Any,1}, ::MXNet.mx.#Activation, args::MXNet.mx.NDArray...) in MXNet.mx
Possible fix, define
  (::MXNet.mx.#kw##Activation)(::Array{Any,1}, ::MXNet.mx.#Activation)
Stacktrace:
 [1] include_from_node1(::String) at ./loading.jl:569
 [2] include(::String) at ./sysimg.jl:14
 [3] process_options(::Base.JLOptions) at ./client.jl:305
 [4] _start() at ./client.jl:371
while loading /home/cokes/.julia/v0.6/MXNet/examples/mnist/lenet.jl, in expression starting on line 10
```

What needs to be changed? Once I learn how, I can update the other examples.

iblis17:
hi @SimonEnsemble ,
I'm on 6d1ba5338c381176ff22e9940430fdcae52c6b52 and Julia 0.6, but cannot reproduce your error.

The error indicate that you have modified file(s) in the working directory (usually `~/.julia/v0.6/MXNet`). `Pkg`(git) reject to checkout master.
```
ERROR: MXNet is dirty, bailing
```

SimonEnsemble:
When I installed MXNet.jl on Ubuntu, I got the following error upon `Pkg.build(MXNet)`:
```
LoadError: failed process: Process(`git submodule update --init`, ProcessExited(1))
```

I then used your solution [here](https://github.com/dmlc/MXNet.jl/issues/263), commenting out the line `git submodule update --init` in `build.jl`, and it built successfully. Could this be the reason I got errors after running the examples?

[...]

PR #287, opened 2017-10-04
build: improve cuda detection

iblis17:
- change ADD_LDFLAGS
        - https://github.com/apache/incubator-mxnet/pull/7856
        - https://github.com/apache/incubator-mxnet/pull/7838

- guessing the CUDA_HOME if user not provides it.

----

@SimonEnsemble
Could you try this patch out with a clean clone?

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -37,7 +37,6 @@ end
 
 HAS_CUDA = false
 HAS_CUDNN = false
-cudnnver = -1
 let cudalib = Libdl.find_library(["libcuda", "nvcuda.dll"], CUDAPATHS)
   HAS_CUDA = !isempty(cudalib) && Libdl.dlopen_e(cudalib) != C_NULL
 end
@@ -53,9 +52,8 @@ end
 if HAS_CUDA  # then check cudnn
   let cudnnlib = Libdl.find_library("libcudnn", CUDAPATHS)
     HAS_CUDNN = !isempty(cudnnlib) && Libdl.dlopen_e(cudnnlib) != C_NULL
-    if HAS_CUDNN
-      # TODO: do more version check?
-      cudnnver = dec(ccall((:cudnnGetVersion, cudnnlib), Csize_t, ()))
+    if HAS_CUDNN && !haskey(ENV, "CUDA_HOME")  # inference `CUDA_HOME`
+      ENV["CUDA_HOME"] = dirname(dirname(cudnnlib))
     end
   end
 end
@@ -63,8 +61,9 @@ end
 if HAS_CUDA
   info("Found a CUDA installation.")
   if HAS_CUDNN
-    info("Found a CuDNN installation (version -> $cudnnver).")
+    info("Found a CuDNN installation.")
   end
+  info("CUDA_HOME -> $(ENV["CUDA_HOME"])")
 else
   info("Did not find a CUDA installation, using CPU-only version of MXNet.")
 end
@@ -170,7 +169,11 @@ if !libmxnet_detected
           ChangeDirectory(_mxdir)
           `git submodule deinit --force .`
           `git fetch`
-          `git checkout $libmxnet_curr_ver`
+          if libmxnet_curr_ver != "master"
+            `git checkout $libmxnet_curr_ver`
+          else
+            `git merge --ff origin/$libmxnet_curr_ver`
+          end
           `git submodule update --init --recursive`
           `git -C mshadow checkout -- make/mshadow.mk`
           `make clean`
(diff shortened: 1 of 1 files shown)


SimonEnsemble:
^ It looks like some checks were not successful?

How can I checkout this particular version?
`Pkg.checkout("MXNet", "iblis17:issue-286")` does not work.

vchuravy:
@SimonEnsemble you need to add as a remote https://github.com/iblis17/MXnet.jl or use Github `hub`

SimonEnsemble:
I tried:
```
Pkg.clone("git://github.com/iblis17/MXnet.jl.git", "issue-286")
Pkg.build("issue-286")
```
and I still get:
```
LoadError: failed process: Process(`make -j8 USE_BLAS=openblas 'MSHADOW_LDFLAGS=-lm /home/cokes/juliav0.6/bin/../lib/julia/libopenblas64_.so'`, ProcessExited(2)) [2]
```
I do have CUDA installed for sure...
`fatal error: cuda.h: No such file or directory`

iblis17:
Curious about where is your cuda and cudnn installation.
On my Linux box, it's /opt/cuda.

iblis17:
Oh, 
you miss this step:
```
Pkg.checkout("issue-286", "issue-286")
```

Pkg.clone only rename the dir and checkout master for you,
so we need to manully checkout this brach of PR.
```

**Comment by SimonEnsemble** [comment]
In my `~/.bashrc`:
```
# for CUDA obviously
export PATH=/usr/local/cuda-8.0/bin:${PATH:+:${PATH}}
export LD_LIBRARY_PATH=/usr/local/cuda-8.0/lib64:${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}
export LD_LIBRARY_PATH=/home/cokes/cudaNN/lib64:${LD_LIBRARY_PATH}
```
Which should I put as my `CUDA_HOME`? Is cudann required for MXNet too?

File: deps/build.jl [hunk 1]
```diff
@@ -53,18 +52,18 @@ end
 if HAS_CUDA  # then check cudnn
   let cudnnlib = Libdl.find_library("libcudnn", CUDAPATHS)
     HAS_CUDNN = !isempty(cudnnlib) && Libdl.dlopen_e(cudnnlib) != C_NULL
-    if HAS_CUDNN
-      # TODO: do more version check?
-      cudnnver = dec(ccall((:cudnnGetVersion, cudnnlib), Csize_t, ()))
+    if HAS_CUDNN && !haskey(ENV, "CUDA_HOME")  # inference `CUDA_HOME`
+      ENV["CUDA_HOME"] = dirname(dirname(cudnnlib))
```
iblis17: YFI: I try to guessing your CUDA_HOME via libcudnn's path

SimonEnsemble: How about this for guessing the path to CUDA?
`filter(x->contains(x, "cuda"), split(ENV["PATH"], ":"))`
I assume everyone with CUDA installed put it in the PATH variable...

SimonEnsemble:
Starting from scratch, I:
```
Pkg.clone("git://github.com/iblis17/MXnet.jl.git", "issue-286")
Pkg.checkout("issue-286", "issue-286")
Pkg.build("issue-286")
```

but it is stuck... unless it takes >15 min to compile...

deps/build.jl
iblis17: 🤔 , it does not set `PATH` and `LD_LIBRARY_PATH` on my machine, but works fine.

BTW, where is your libcublas
```
julia> Libdl.dlpath("libcublas")                                                                                                                  
"/opt/cuda/lib64/libcublas.so
```
seesm you have different installation path for CUDA and cuDNN,
maybe I should determin path from cuBLAS?

iblis17:
> but it is stuck... unless it takes >15 min to compile...

yeah, on Travis CI, It takes > 20 min

iblis17:
You can check if something eat up your CPUs via `top` or not.

iblis17:
> Which should I put as my CUDA_HOME? Is cudann required for MXNet too?

* In your case, I expect `CUDA_HOME` is `/usr/local/cuda-8.0/`, `CUDA_HOME` should contains `include`, `bin` and `lib64` in it.
* cuDNN is optional.

SimonEnsemble:
@iblis17 It compiles and the example on the README finally runs without errors!

So please do merge into main MXNet.
Another alternative is to just have users install MXNet from here:
https://mxnet.incubator.apache.org/get_started/install.html
and then always use
export MXNET_HOME before installation.
But when I do that, while it compiles, there are errors that arise in the examples, so it seems that MXNet.jl is predicated upon a different version of MXNet.

iblis17:
🎉 

Let's wait for @vchuravy 's review.

>Another alternative is to just have users install MXNet from here:
https://mxnet.incubator.apache.org/get_started/install.html
and then always use
export MXNET_HOME before installation.
But when I do that, while it compiles, there are errors that arise in the examples, so it seems that MXNet.jl is predicated upon a different version of MXNet.

Master branch of MXNet.jl requires master branch of upstream mxnet.
Maybe checkout the stable release MXNet.jl can works with the prebuilt binary version. (but I don't try it yet)

hmm ... https://github.com/dmlc/MXNet.jl/releases
the last release is for upstream mxnet 0.10.0 😱 ...

SimonEnsemble:
Now I'm installing MXNet.jl on CentOS.
I have a working cuda installation in `/usr/local/apps/cuda/cuda-8.0`.

On master (via `Pkg.checkout("MXNet")`), upon `Pkg.build("MXNet")`, it does not detect the CUDA installation.

I then:
`ENV["CUDA_HOME"] = "/usr/local/apps/cuda/cuda-8.0"; Pkg.build("MXNet")`
and still get the message:
`INFO: Did not find a CUDA installation, using CPU-only version of MXNet.`

What can I do to troubleshoot?

pluskid:
@SimonEnsemble have you tried to use [manual installation](http://dmlc.ml/MXNet.jl/latest/user-guide/install/#Manual-Compilation-1) that compile libmxnet separately and point to the compiled library?

CorySimon:
@pluskid Yes, I've successfully built MXNet.jl pointing to a version of MXNet that I compiled myself. But, then the examples in MXNet.jl do not even run... See here:
https://github.com/dmlc/MXNet.jl/issues/286

Issue #170, opened 2017-01-16
Define broadcast for NDArray


vchuravy:
Right now we are getting a lot of deprecated syntax warnings on v0.6:

 ```
WARNING: deprecated syntax "function .+(...)".
Use "function Base.broadcast(::typeof(+), ...)" instead.
```

In general we should implement `broadcast/broadcast!` over NDArrays/Symbols exploiting syntactic loop fusion that 0.6 (and v0.5 to some extend) provides. 

Related issues:
#109, #152 

Custom operators is something I only plan to implement on v0.5 and to run them on the GPU with CUDAnative will require v0.6

Arkoniak:
From this ref: https://discourse.julialang.org/t/how-to-handle-deprecations-of-new-broadcast-functions/1287 it looks like it can be solved only after corresponding `@compat` macro is going to be added in `Compat.jl`, or write custom macro. Are there any other options?

Arkoniak:
Upd: There as in issue now https://github.com/JuliaLang/Compat.jl/issues/308

vchuravy:
We can definitely start thinking how we are going to implement broadcasting in the long run, but you are right this will depend on Compat providing a way of supporting both 0.5 and 0.6

A way forward would be to implement broadcasting over NDArrays by copying the data to the CPU and let Julia perform loop-fusion, later on we can use a similar approach tho GPUArrays.jl to offload kernel functions to the GPU.

More complicated will be supporting SymbolicNodes... ideally we would do something very similar with NativeSymbols, but that is currently not possible and using NativeModules might be to invasive and not good for performance.

pluskid:
Haven't been following Julia broadcast syntax changes yet. It seem to be more complicated than simple syntax sugar that turns `a .* b` into something like `broadcast(*, a, b)`, as it also deals with more complicated expressions?

But if we could do it that way by properly delegating to the libmxnet broadcasting operator calls (symbolic or ndarray), then I guess it won't be very complicated (such as launching GPU kernels manually).

vchuravy:
There is a good overview of what the broadcast syntax changes entail: http://julialang.org/blog/2017/01/moredots
The problem is that `X .+ Y .* Z` is lowered to `broadcast((x,y,z)->x+y*z, X, Y, Z)` and works for arbitrary Julia functions.

vchuravy:
I chatted with @denizyuret about this and he pointed me to a few resources how other projects solved this.

[...]

PR #292, opened 2017-10-08
ndarray: implement rdiv

iblis17:
e.g.

```julia
1 ./ mx.NDArray(Float32[1 2; 3 4])
```

### DIFF
--- a/src/ndarray.jl
+++ b/src/ndarray.jl
@@ -681,25 +681,42 @@ function div_from!(dst :: NDArray, arg :: Union{Real, NDArray})
   end
 end
 
+"""
+Elementwise division of NDArray
+"""
+div(x::NDArray, y::NDArray) = _div(x, y)
+div(x::NDArray, s::Real) = _div_scalar(x, scalar=s)
+div(s::Real, x::NDArray) = _rdiv_scalar(x, scalar=s)
+
 import Base: /
 """
     ./(arg0 :: NDArray, arg :: Union{Real, NDArray})
 
 Elementwise dividing an `NDArray` by a scalar or another `NDArray` of the same shape.
 """
-@compatdot function Base.broadcast(::typeof(/), arg0 :: NDArray, arg :: Union{Real, NDArray})
-  ret = copy(arg0, context(arg0))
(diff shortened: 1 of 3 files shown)


src/ndarray.jl
```diff
@@ -681,25 +681,42 @@ function div_from!(dst :: NDArray, arg :: Union{Real, NDArray})
...
-  arg0 ./ arg
-end
+/(arg0 :: NDArray, arg :: Real) = div(arg0, arg)
+
+"""
+    /(arg0 :: Real, arg :: NDArray)
+
+Elementwise Divide a scalar by an `NDArray`.
+"""
+/(arg0 :: Real, arg :: NDArray) = div(arg0, arg)
```
vchuravy: I am not a fan of this since it it no longer Julia semantics

iblis17: hmm ... I misunderstood that julia 0.5 need it for broadcasting.
(maybe I should not write code in early morning :p

iblis17: Oh... we still need it because of Julia 0.6's broadcasting fusion

```julia
julia> expand(:(1 ./ mx.zeros(4)))
:($(Expr(:thunk, CodeInfo(:(begin 
        $(Expr(:thunk, CodeInfo(:(begin 
        global ##5#6
        const ##5#6
        $(Expr(:composite_type, Symbol("##5#6"), :((Core.svec)()), :((Core.svec)()), :(Core.Function), :((Core.svec)()), false, 0))
        return
    end))))
        $(Expr(:method, false, :((Core.svec)((Core.svec)(##5#6, Core.Any), (Core.svec)())), CodeInfo(:(begin 
        return 1 / #temp#
    end)), false))
        #5 = $(Expr(:new, Symbol("##5#6")))
        SSAValue(0) = #5
        SSAValue(1) = ((Core.getfield)(mx, :zeros))(4)
        return (Base.broadcast)(SSAValue(0), SSAValue(1))
    end)))))
```

vchuravy: We need to solve #170. Before we can do this.

PR #293, opened 2017-10-08
ndarray: elementwise power

iblis17:
```julia
x.^2

2.^x

x.^y
```


src/ndarray.jl
```diff
@@ -701,6 +701,50 @@ function /(arg0 :: NDArray, arg :: Real)
...
+"""
+    ^(x::NDArray, y::NDArray)
+    ^(x::NDArray, s::Real)
+    ^(s::Real, x::NDArray)
+
+Elementwise power
+"""
+^(x::NDArray, y::NDArray) = power(x, y)
+
+^(x::NDArray, s::Real) = power(x, s)
```
vchuravy: I don't think this is the semantic we want. In Julia we would get A^2 = A*A which is matrix multiply and not the elementwise power. Now the question is what to do for tensors instead of matrices and if NDArrays are mathematical constructs or containers. In any case I would prefer to only define the broadcasting operation since that is well-defined.

iblis17: I think it's same story: https://github.com/dmlc/MXNet.jl/pull/292/files#r143826115

pluskid:
I agree with @vchuravy here. I think we should remove the non-elementwise implementation here or even implement it as throwing an error to warn the user. (same for rdiv).

iblis17:
about dot-call, I will refine this PR and rdiv after #300 finished

iblis17:
I think this PR is ready for review.

PR #294, opened 2017-10-09
ndarray: getindex/setindex! linear indexing

iblis17:
```julia
x = mx.zeros(2, 5)
x[5] = 42
```


src/ndarray.jl
```diff
@@ -396,13 +417,36 @@ function getindex(arr :: NDArray, ::Colon)
...
+
+  idx -= 1
+  offsets = size_[1:end-1] |> reverse ∘ cumprod ∘ collect
+  handle = arr.handle
+  for offset ∈ offsets
+    handle = _at(handle, idx ÷ offset)
+    idx %= offset
+  end
+
+  _at(handle, idx) |> MX_NDArrayHandle |> x -> NDArray(x, arr.writable)
```
vchuravy: It is kinda weird to return a one element array and also not common in Julia. The problem is if we convert it to a Julia value at this point we impose a synchronisation barrier.

iblis17: I agree with the weirdness, Python provide an API `nd.asscalar()` for conducting read value.
I think in Julia, we can have `first` for reading it.

looks like this
```julia
x[42] |> first  # do synchronisation and get the value
```

and keep `x[42]` returning an `NDArray`

iblis17: I can send a PR for `first` api. If you agree with that design

pluskid: +1 for the `first` design. I think we should make it explicit for users that an `NDArray` is not some transparent array, and it could potentially be on GPU and could potentially be slow to do indexing into every single element of it.

BTW: I feel I'm pretty out-dated about the latest Julian things when I see the `|>` and `∘` operators. :D

iblis17: `first` implemented @ 62bd213

NEWS.md
```diff
@@ -26,6 +26,26 @@
      2.0  4.0
     ```
 
+* `NDArray` `getindex`/`setindex!` linear indexing support and `first` for extracting scalar value. (#TBD)
```
pluskid: `#TBD`?

pluskid:
Thanks! LGTM. Can you resolve the conflicts in NEWS.md?

NEWS.md
iblis17: (at first I wrote this patch, the issue number was still unknown)
I will send a PR to sort them out for releasing.

iblis17:
resolved

PR #297, opened 2017-10-10
ndarray: implement `fill`

iblis17:
It serve as an API corresponds to Python's `mx.nd.full()`


src/ndarray.jl
```diff
@@ -703,6 +703,27 @@ end
 
 
 """
+    fill(x, arr::NDArray, ctx=cpu())
+    fill(x, dims, ctx=cpu())
+    fill(x, dims...)
+
+Create an `NDArray` filled with the value `x`, like `Base.fill`.
+"""
+function fill(x, arr::NDArray, ctx::Context=cpu())
```
vchuravy: Drop the `ctx` argument here, and maybe call this `fill!`?

iblis17:
ready for review

PR #299, opened 2017-10-26
Fix `@mx.chain` var reference in macro

### LINKED ISSUE #298: mlp.jl: accept external arguments?
When using the neural net factory, 
 `mx.MLP([20, 10, 6])`
works fine to set up part of a chain. However, I would like to include this into a more general function that sets up a net taking the number of layers and the neurons per layer as arguments. That is, I would like to be able to use something like
 `mx.MLP(netconfig)`
where netconfig is previously defined to be `netconfig=[20, 10, 6]` or some other description which can easily be passed as an argument. 

This sort of thing used to work, before the changes that were done to get MXNet.jl working with julia 0.6.x. I haven't figured out how to do it with the current master version. Any help would be appreciated.

iblis17:
fix #298


iblis17:
well, Appveyor always hits the download issue. :/

pluskid:
It seems Travis always timeout for Mac & Julia 0.5

PR #300, opened 2017-10-30
ndarray: intro TakingBroadcastSeriously and broadcast unfusing

iblis17:
Ref: https://github.com/dmlc/MXNet.jl/issues/170#issuecomment-330080485

Could we drop 0.5 support? or just self-shipping the source of [TakingBroadcastSeriously](https://github.com/MikeInnes/TakingBroadcastSeriously.jl)?


vchuravy:
Dropping 0.5 at this point should be fine.

pluskid:
+1 for dropping 0.5 support

iblis17:
Will add some commits later in this PR. :)

iblis17:
CI failures should be fixed in #301

PR #306, opened 2017-11-08
Reexporting NDArray and SymbolicNode

iblis17:


pluskid:
Thanks! LGTM, but why are we exporting `NDArray` and `SymbolicNode` as oppose to keeping everything under `mx`? Is there a particular scenario that you have in mind that this will make it easier?

iblis17:
This is an initial work for reexporting.
Feel free to list others here you want to export. :)

iblis17:
I added tons of exports.
Good to go?

pluskid:
travis error-ed, otherwise LGTM.

iblis17:
ok, i fixed the tests.

PR #307, opened 2017-11-08
Fix deprecations

iblis17:


iblis17:
I take this patch from my personal fork: https://github.com/iblis17/MXNet.jl/pull/1

PR #311, opened 2017-11-09
ndarray: regularize fill! API

iblis17:
In Base, it's

        fill!(A, x)


pluskid:
Can you do a quick search through the code to check if it breaks anything? (optimizers, initializers, etc.)

iblis17:
I add `fill!` in this PR, so I assume that nobody is using it . 
https://github.com/dmlc/MXNet.jl/pull/297

pluskid:
Got it!

PR #326, opened 2017-11-16
Bump upstream to v0.12.1 on Linux/macOS

iblis17:
* there is still no windows binary at this moment.


iblis17:
travis failure is hitting time limit (50 mins) on macOS build.

pluskid:
It seems that travis CI fails half of the time on OS X build for timeout. Wonder if there is anything we can do...

iblis17:
The 50 mins of limitation is a hard limit for open source project: https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts

Or... is there nightly libmxnet binary for macOS ?

pluskid:
I don't think we have nightly libmxnet build for macOS. Is the building part taking up the most of the 50-min limit? Would some of the [caching mechanism](https://docs.travis-ci.com/user/caching/) help?

iblis17:
https://github.com/dmlc/MXNet.jl/blob/9efedbccc098a3b1eb1288ba6e51b61d8d3e169b/deps/build.jl#L71-L73

I can tweak the # of make job.

iblis17:
> Would some of the caching mechanism help?

Oh, will try it.

PR #331, opened 2017-11-18
parametric NDArray

iblis17:


src/executor.jl
```diff
@@ -5,19 +5,25 @@ An executor is a realization of a symbolic architecture defined by a `SymbolicNo
...
 be carried out with an executor.
 """
-mutable struct Executor
+const VecOfNDArray = AbstractVector{<:NDArray}
+mutable struct Executor{A<:VecOfNDArray,
+                        B<:VecOfNDArray,
+                        G<:AbstractVector{<:Union{Void,NDArray}},
+                        O<:VecOfNDArray,
+                        D<:Dict{Symbol,<:NDArray},
+                        E<:Dict{Symbol,<:NDArray}}
```
iblis17: This is terrible, any better way to deal with abstract container type?
Or we need to redesign `Executor`?

iblis17:
with parametric NDArray, constructing `Dict` will easily get `Dict{Symbol, Any}`. Is this ok for us?
Before parametric, we have lots of functions (like `bind`) constrain some input type as `Dict{Symbol, NDArray}`.

```julia
julia> Dict(:x => x, :y => y)                                                                                                                     
Dict{Symbol,Any} with 2 entries:                                                                                                                  
  :y => 6 mx.NDArray{Float32,1} @ CPU0:…                                                                                                          
  :x => 2×3 mx.NDArray{Float32,2} @ CPU0:…
```

pluskid:
hmm, I would like to avoid being resolved to Any if possible.

iblis17:
Actually, same story on `Base.Array`
```julia
julia> x = [1,2]
2-element Array{Int64,1}:
 1
 2

julia> y = [1 2; 3 4]
2×2 Array{Int64,2}:
 1  2
 3  4

julia> Dict(:x => x, :y => y)
Dict{Symbol,Any} with 2 entries:
  :y => [1 2; 3 4]
  :x => [1, 2]
```

src/metric.jl
```diff
@@ -235,31 +221,34 @@ Calculates the mean squared error regression loss.
 Requires that label and prediction have the same shape.
 """
 
-mutable struct MSE <: AbstractEvalMetric
-  mse_sum  :: Vector{NDArray}
+mutable struct MSE{T<:NDArray} <: AbstractEvalMetric
+  mse_sum  :: Vector{T}
```
iblis17: @vchuravy I changed `MSE` like this, but I got confused that how does user initialize a `MSE` instance?
In the example of regression, it just creates a instance via `MSE()`.
And...what's your design of struct `MSE` if we have parametric `NDArray`?

vchuravy: Since this is for error reporting only it should simply default to `Float64` or `Float32`.
The array is simply to not block during `_update_single_output`.

iblis17: > Since this is for error reporting only it should simply default to `Float64` or `Float32`.

okay, I will init them with `MX_float` (`Float32`) by default.

> The array is simply to not block during `_update_single_output`.

I do not get what you mean.

iblis17:
This PR is ready for review now.

iblis17:
good to go ?

PR #336, opened 2017-11-20
appveyor: enable integration test

iblis17:


iblis17:
omg, it failed.

pluskid:
hmm, it seems it is not reaching the expected performance under windows? First thing might be do some regression test on some older version. Maybe easier if we have local windows machines, but I don't think any one of us actually have.

PR #339, opened 2017-11-22
Update regression-example.jl

mcreel:
Fix the call to scatter plot, in the last line. It is commented out, but if a user uncomments it to see a plot of the fit, the fix is needed.


iblis17:
👍

PR #342, opened 2017-11-24
example: fix batchsize config in regression-example

iblis17:
@mcreel could you review this?


mcreel:
Sorry, I missed the message asking for a review. Thanks for giving it the update.

PR #344, opened 2017-11-24
build: opencv auto detection

iblis17:


pluskid:
Both Mac & Linux building fails, though seem to be due to different reasons.

PR #351, opened 2017-12-01
ndarray: implement endof

iblis17:


pluskid:
Is this used for iteration? I think in the Python side the default behavior is to iterate over the 0-th dimension (the last dimension for Julia). While I think it is OK to have our API diverging from the Python because we already have a lot of Julia flavor API, but I think at some point we should add a section to the introduction or some relevant part of our doc to emphasize this.

iblis17:
`endof` is for the `x[end]`, e.g.
```julia
julia> x
4 mx.NDArray{Int64,1} @ CPU0:
 1
 2
 3
 4

julia> x[end]
1 mx.NDArray{Int64,1} @ CPU0:
 4
```

iblis17:
FYR: https://docs.julialang.org/en/latest/manual/interfaces/#Indexing-1

pluskid:
OK, thanks!

Issue #353, opened 2017-12-02
ndarray: just let libmxnet do its own type convertion on `*_scalar`


iblis17:
In our codebase, we always do `mx._mul_scalar(x, scalar = convert(..., π))`

but this looks fine.
```julia
julia> x
3 mx.NDArray{Int64,1} @ CPU0:
 1
 2
 3

julia> mx._mul_scalar(x, scalar = π)
3 mx.NDArray{Int64,1} @ CPU0:
 3
 6
 9
```

I will send PR one-by-one
check list:
- [x] `_plus_scalar`： #360 
- [x] `_minus_scalar`: #354 
- [x] `_mul_scalar`: #356 
- [x] `_div_scalar`: #357 
- [x] `_rdiv_scalar`: #358

iblis17:
All PRs sent. Please review.

PR #357, opened 2017-12-02
ndarray: type convertion of _div_scalar

iblis17:
Ref: #353

- and handle the case of integer NDArray divided by zero


src/ndarray.jl
```diff
@@ -656,13 +656,21 @@ Matrix (2D NDArray) multiplication.
...
+    _div_scalar(dst, scalar = arg, out = dst)
   else
     _div(dst, arg, out = dst)
   end
+  dst
+end
+
+function div_from!(dst::NDArray{T}, arg::Real) where {T<:Integer}
+  @assert dst.writable
+  @assert(round(T, arg) != zero(T), "Integer divided by zero")
```
pluskid: It seems Julia built-in array throws `InexactError` instead of assertion error when assigning out of bound floating point to integer arrays. Shall we follow the convention?

iblis17: In Julia's Array, dividing-by-zero will return `Inf`,
 and the type of Array become `Array{Float64}`
```julia
julia> [1, 2, 3] / 0                                                                                                                    
3-element Array{Float64,1}:                                                                                                             
 Inf                                                                                                                                    
 Inf                                                                                                                                    
 Inf
```

But, if I try to do this in libmxnet. libmxnet just raise segfault, and crash julia process.
That's why I try to do this kind of protection.
```julia
julia> using MXNet                                                                                                                      
INFO: Recompiling stale cache file /home/iblis/.julia/lib/v0.6/MXNet.ji for module MXNet.                                               
                                                                                                                                        
julia> mx.NDArray([1, 2, 3]) / 0                                                                                                        
zsh: segmentation fault (core dumped)  julia
```

iblis17:
ok to go ?

pluskid:
Sorry, I missed the the previous conversation. What I meant is this:
```julia
julia> x = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> x ./= 0
ERROR: InexactError()
Stacktrace:
 [1] convert(::Type{Int64}, ::Float64) at ./float.jl:679
 [2] setindex! at ./multidimensional.jl:300 [inlined]
 [3] macro expansion at ./broadcast.jl:154 [inlined]
 [4] macro expansion at ./simdloop.jl:73 [inlined]
 [5] macro expansion at ./broadcast.jl:147 [inlined]
 [6] _broadcast!(::##1#2, ::Array{Int64,1}, ::Tuple{Tuple{Bool}}, ::Tuple{Tuple{Int64}}, ::Array{Int64,1}, ::Tuple{}, ::Type{Val{0}}, ::CartesianRange{CartesianIndex{1}}) at ./broadcast.jl:139
 [7] broadcast_c! at ./broadcast.jl:211 [inlined]
 [8] broadcast!(::Function, ::Array{Int64,1}, ::Array{Int64,1}) at ./broadcast.jl:204
```
Julia raise about inexact error when converting the floating `Inf` to integers. I'm also fine with the assertion error. Feel free to merge it.

iblis17:
About the InexactError,
I want to rely on the type conversion of libmxnet.
Libmxnet is quite different from Julia's conversion policy.
For example, in Julia, an int array `A` multiply a float will return a float array,
but in libmxnet, it all the return type of NDArray is fixed, cannot be altered.
So... IMO, I'm not going to support/raise InexactError from NDArray.
That's why I opened this series of issue.
```julia
julia> x=[1, 2, 3]                                                                                                                                
3-element Array{Int64,1}:                                                                                                                         
 1                                                                                                                                                
 2                                                                                                                                                
 3                                                                                                                                                
                                                                                                                                                  
julia> x[1] = 1.1                                                                                                                                 
ERROR: InexactError()                                                                                                                             
Stacktrace:                                                                                                                                       
 [1] convert(::Type{Int64}, ::Float64) at ./float.jl:679                                                                                          
 [2] setindex!(::Array{Int64,1}, ::Float64, ::Int64) at ./array.jl:583
```

```julia
julia> x = NDArray([1,2,3])                                                                                                                       
3 mx.NDArray{Int64,1} @ CPU0:                                                                                                                     
 1                                                                                                                                                
 2                                                                                                                                                
 3                                                                                                                                                
                                                                                                                                                  
julia> x[1] = 1.1                                                                                                                                 
1.1                                                                                                                                               
                                                                                                                                                  
julia> x                                                                                                                                          
3 mx.NDArray{Int64,1} @ CPU0:                                                                                                                     
 1                                                                                                                                                
 2                                                                                                                                                
 3
```

pluskid:
OK, I'm on board. :)

PR #361, opened 2017-12-03
ndarray: make trigonometric functions dot-call only

iblis17:
- Also, renamed arc* funcs to a* to keep consistent with Base
- deprecated non-dot-call


iblis17:
I make them dot-call only, because almost all the functions which can work with single value will pop depwarn on array since Julia 0.6 if user invoke without dot-call .
```julia
julia> a = [1, 2, 3]                                                                                                                              
3-element Array{Int64,1}:                                                                                                                         
 1                                                                                                                                                
 2                                                                                                                                                
 3                                                                                                                                                
                                                                                                                                                  
julia> sin(a)                                                                                                                                     
WARNING: sin(x::AbstractArray{T}) where T <: Number is deprecated, use sin.(x) instead.                                                           
Stacktrace:                                                                                                                                       
 [1] depwarn(::String, ::Symbol) at ./deprecated.jl:70                                                                                            
 [2] sin(::Array{Int64,1}) at ./deprecated.jl:57                                                                                                  
 [3] eval(::Module, ::Any) at ./boot.jl:235                                                                                                       
 [4] eval_user_input(::Any, ::Base.REPL.REPLBackend) at ./REPL.jl:66                                                                              
 [5] macro expansion at /home/iblis/.julia/v0.6/Revise/src/Revise.jl:775 [inlined]                                                                
 [6] (::Revise.##17#18{Base.REPL.REPLBackend})() at ./event.jl:73                                                                                 
while loading no file, in expression starting on line 0                                                                                           
3-element Array{Float64,1}:                                                                                                                       
 0.841471                                                                                                                                         
 0.909297                                                                                                                                         
 0.14112
```

PR #363, opened 2017-12-05
ndarray: add Base.ones(::NDArray) and Base.zeros

iblis17:
For creating NDArray with same type and dims


iblis17:
Any objection of this?

pluskid:
LGTM

PR #391, opened 2017-12-28
ndarray: refine signature of `clip`

iblis17:
Before: `clip(x, a_min = -4, a_max = 4)`
After: `clip(x, -4, 4)`

The `a_min` and `a_max` is a required but they are keyword arguments.
So this commit change them to positional arguments and make sure user will
fill them.


iblis17:
will implement `Base.clamp` after this PR merged

PR #394, opened 2017-12-30
opt: rename keyword args of NadamScheduler

iblis17:
* Before

```julia
NadamScheduler(; mu0 = 0.99, delta = 0.004, gamma = 0.5, alpha = 0.96)
```

* After

```julia
NadamScheduler(; μ₀ = 0.99, δ = 0.004, γ = 0.5, α = 0.96)
```


iblis17:
covered in #396

PR #396, opened 2017-12-31
Optimizer module overhaul

iblis17:
will sort out summary later...


iblis17:
Ready for review.
I list changes in NEWS.md:
 https://github.com/dmlc/MXNet.jl/pull/396/files#diff-8312ad0561ef661716b48d09478362f3R263

The motivation of this PR is building more elegant APIs than Python (thanks to good Unicode support of Julia's REPL and editor plugin),
and I'm paving the way for porting Python's Gluon.

pluskid:
Thanks for the efforts.

* I'm OK with most of the renaming to unicodes.
* I'm a bit against using `\eta` for learning rate, `\mu` for momentum and `\lambda` for weight decay. Although it seems to be commonly used, but some papers use different symbols and could be a bit confusing. But those are still OK if you insist.
* I'm strictly against using `∇c` for gradient clip and `∇r` for gradient rescaling. Those are just super confusing. Because `∇` typically means taking gradient with respect to something or similar things.

iblis17:
> - I'm a bit against using `\eta` for learning rate, `\mu` for momentum and `\lambda` for weight decay. Although it seems to be commonly used, but some papers use different symbols and could be a bit confusing. But those are still OK if you insist.
> - I'm strictly against using `∇c` for gradient clip and `∇r` for gradient rescaling. Those are just super confusing. Because `∇` typically means taking gradient with respect to something or similar things.

I agree the your point of view on gradient clipping and rescaling.

I want to hear what naming you want. We can list all of them here, then vote.

pluskid:
Unfortunately, I don't have better naming suggestion apart from the more verbose `grad_clip` and `grad_scale`.

iblis17:
Add another options:
- `clip`
- `rescale`

iblis17:
Is it okay to omit the `grad_` prefix, given that optimizor is related to gradient stuff?

iblis17:
I did the renaming, please check it out.

iblis17:
good to go?

pluskid:
Sorry for the late reply. I still prefer with `grad_` prefix, but I do not have strong objection to the current naming as it is the options for optimizers as you mentioned (and we probably are not going to have scales for momentum and other quantities). Please feel free to merge it.

iblis17:
Thanks a lot.

About the `grad_` prefix, there is a way to add aliases in Julia 0.7 (https://github.com/JuliaLang/julia/pull/24960). Once we drop the support of Julia 0.6, I can add the prefix as aliases.

pluskid:
Alias sounds good!

